PR #5, opened 2019-06-07
WIP sketch out the skeleton

raulk:
Doesn't compile. Placeholder PR where I'll keep working!


src/main/kotlin/io/libp2p/core/protocol/Negotiator.kt
```diff
@@ -0,0 +1,135 @@
+package io.libp2p.core.protocol
+
+import io.libp2p.core.events.ProtocolNegotiationFailed
```
Nashatyrev: @raulk you forgot to add those event classes

raulk: Pushed!

PR #6, opened 2019-06-11
Implement MultiAddr

Nashatyrev:
There is still a dependency https://github.com/multiformats/java-multiaddr to address P2P protocol. 
Also there is a bug with `Base32.encode()` there which affects `ONION` protocol
Should we address those issues now or just create a couple of issues and leave it as is for now?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,7 @@ plugins {
 
 repositories {
     mavenCentral()
+    maven("https://jitpack.io")
 }
 
 dependencies {
@@ -21,6 +22,8 @@ dependencies {
     compile("com.google.guava:guava:27.1-jre")
     compile("org.bouncycastle:bcprov-jdk15on:1.61")
     compile("org.bouncycastle:bcpkix-jdk15on:1.61")
+    compile("com.github.multiformats:java-multiaddr:v1.3.1")
+
 
     testCompile("org.junit.jupiter:junit-jupiter-api:5.4.2")
     testCompile("org.junit.jupiter:junit-jupiter-params:5.4.2")
--- a/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
+++ b/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
@@ -4,6 +4,7 @@ import io.ipfs.multiaddr.Base32
 import io.ipfs.multiaddr.Protocol.LENGTH_PREFIXED_VAR_SIZE
 import io.libp2p.core.types.readUvarint
 import io.libp2p.core.types.toByteArray
+import io.libp2p.core.types.toByteBuf
 import io.libp2p.core.types.writeUvarint
 import io.netty.buffer.ByteBuf
 import java.net.Inet4Address
@@ -32,7 +33,7 @@ enum class Protocol(val code: Int, val size: Int, val typeName: String) {
     },
     IPFS(421, LENGTH_PREFIXED_VAR_SIZE, "ipfs"),
     HTTPS(443, 0, "https"),
-    ONION(444, 80, "onion"),
+    ONION(444, 96, "onion"),
     QUIC(460, 0, "quic"),
     WS(477, 0, "ws"),
     P2PCIRCUIT(290, 0, "p2p-circuit"),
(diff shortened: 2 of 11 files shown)


src/main/kotlin/io/libp2p/core/types/Uvarint.kt
```diff
@@ -2,21 +2,24 @@ package io.libp2p.core.types
 
 import io.netty.buffer.ByteBuf
 
-fun ByteBuf.writeUvarint(value: Long) {
+fun ByteBuf.writeUvarint(value: Int): ByteBuf = writeUvarint(value.toLong())
```
raulk: Nice!

src/main/kotlin/io/libp2p/core/types/ByteArrayExtensions.kt
```diff
@@ -0,0 +1,6 @@
+package io.libp2p.core.types
+
+fun ByteArray.toHex() = this.joinToString(separator = "")
```
raulk: Extension functions are so nice ;-)

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -0,0 +1,157 @@
...
+    HTTP(480, 0, "http");
+
+    val encoded: ByteArray = encode(code)
+
+    private fun encode(type: Int): ByteArray =
+        byteBuf(4).writeUvarint(type).toByteArray()
+
+    open fun isPath() = false
+
+    fun addressToBytes(addr: String): ByteArray =
```
raulk: Eventually I'd like to get to a point where we use some form of transcoder functions + generics. So each protocol would be associated with a `data class` that encapsulates its fields/values, strongly typed. And we'd register the protocols using reified generics (super nice in Kotlin, I believe they translate to `Class<?>` parameters in plain Java). Then you can do something like:

```
val ma : Multiaddr = foo
val ipv4 : IPv4 = ma.first<IPv4>()
println(ipv4.address)
```

But let's leave that for later!

build.gradle.kts
```diff
@@ -21,6 +22,8 @@ dependencies {
     compile("com.google.guava:guava:27.1-jre")
     compile("org.bouncycastle:bcprov-jdk15on:1.61")
     compile("org.bouncycastle:bcpkix-jdk15on:1.61")
+    compile("com.github.multiformats:java-multiaddr:v1.3.1")
```
raulk: Where is this actually used? I think I missed it.

src/main/kotlin/io/libp2p/core/types/Uvarint.kt
```diff
@@ -2,21 +2,24 @@ package io.libp2p.core.types
...
+
+fun ByteBuf.writeUvarint(value: Long): ByteBuf {
     var v = value
     while (v >= 0x80) {
         this.writeByte((v or 0x80).toInt())
         v = v shr 7
     }
     this.writeByte(v.toInt())
+    return this
```
raulk: Nice, this respects the general contract of ByteBuf!

raulk:
Gonna take the liberty of merging this into the `sketch` branch, to unblock #7 which will create a few conflicts but expands on the Uvarint tests, so definitely want to merge that.

build.gradle.kts
Nashatyrev: `Cid` class: https://github.com/raulk/jvm-libp2p-minimal/blob/b60fe22ec05dab0b2b495e6bb0fbe66c0578837b/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt#L70
`Base32` class: https://github.com/raulk/jvm-libp2p-minimal/blob/b60fe22ec05dab0b2b495e6bb0fbe66c0578837b/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt#L83

@raulk We can exclude IPFS & ONION protocols for now and remove that dependency

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
Nashatyrev: Yup, got you!
BTW, I didn't try adding too many funcs to `Multiaddr`. Will be adding them on demand later while implementing other parts

raulk:
@Nashatyrev BTW – for stuff we know needs doing, I think it's fair to add TODO comments in code at this point. We're _that_ hacky at this stage 😛

PR #11, opened 2019-06-25
continue sketching the system

raulk:
This PR is against an old commit in the `sketch` branch -- that's why it can't be merged immediately. If we agree with the design direction being postulated, I'm happy to rebase.

* Initial spike for a configuration DSL.
* Introduce the concept of a `ProtocolBinding`, which returns a future to a controller object for that protocol (can also be Void).
* `SecureChannel` and `StreamMuxer` are both specialised versions of `ProtocolBinding`, returning a `SecureSession` (which exposes identity information) and `MuxerSession` (which provides capabilities to open/close/manage streams).
* Add an `AddressBook` interface.
* Start modelling a `Connection` class, which wraps a `Channel` representing the connection, and provides accessors for capabilities (muxing and secure channel), so far.
* Sketch out the `Transport` interface and the logic for listening and dialling.


src/main/kotlin/io/libp2p/core/transport/tcp/TcpTransport.kt
```diff
@@ -0,0 +1,66 @@
...
+            if (!it.isSuccess) {
+                ret.completeExceptionally(it.cause())
+            }
+            upgrader.establishMuxer(it.channel()).addListener(muxerListener)
+        }
+
+        // TODO: add an overarching timeout covering all upgrades.
+        client.connect().addListener(secureChannelListener)
+        return ret
+    }
```
Nashatyrev: I would do something in `CompletableFuture` style here. Like: 
```kt
        return client.connect().toCompletableFuture()
            .thenCompose { upgrader.establishMuxer(it).toCompletableFuture() }
            .thenApply { Connection(it) }

[...]

fun ChannelFuture.toCompletableFuture(): CompletableFuture<Channel> = ...
```

src/main/kotlin/io/libp2p/core/transport/Transport.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Stops the transport entirely, closing down all ongoing connections, outbound dials, and listening endpoints.
+     */
+    fun close(): CompletableFuture<Void>
+
+    /**
+     * Makes this transport listen on this multiaddr. The future completes once the endpoint is effectively listening.
+     */
+    fun listen(addr: Multiaddr): CompletableFuture<Void>
```
Nashatyrev: How incoming connections are handled? It would be perfect to return here something like
```kt
Publisher<Connection>  // RX style

or

Channel<Connection>  // Kotlin style
```

Nashatyrev: Or with the Netty style it could be 
```kt
fun listen(addr: Multiaddr, chInit: ChannelInitializer): CompletableFuture<Void>
```

raulk: So I don't think we'd return a specific connection here. This method begins _listening_ on a particular multiaddr. In Netty terms, it probably starts a `ServerBootstrap`. We need to attach a handler to the server to process `channelActive` events (incoming sockets) and upgrade them (negotiate the secure channel and multiplexing).

At that point the `Transport` will have a `Connection` and it'll need to register it in the `Network` by calling `Network#addConnection()`, so we need to pass a reference to the `Network` into the `Transport` (maybe in the constructor for now?).

We also need to be able to remove connections when the `channelInactive` event fires.

Note that protocols and user-land code can subscribe to `Connected` and `Disconnected` notifications at the libp2p layer. See a notifee in action here: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/notif.go, and the way it registers itself with the Network: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/dht.go#L90

We should probably NOT use this exact approach, but rather the local event bus, which we landed recently in go-libp2p: https://github.com/libp2p/go-libp2p/issues/653.

src/main/kotlin/io/libp2p/core/transport/tcp/TcpTransport.kt
raulk: Ahhh that's nice! Feel free to submit a PR for that.

raulk:
@nashatyrev thanks for the review. I’m ok to merge this, I just don’t have bandwidth to fix the conflicts. Do you have any spare cycles? Thanks!

src/main/kotlin/io/libp2p/core/protocol/ProtocolBinding.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Matching heuristic to be used on inbound negotiations. If positive, it leads to activation of this protocol.
+     */
+    val matcher: ProtocolMatcher
+
+    /**
+     * Returns initializer for this protocol on the provided channel, together with an optional controller object.
+     */
+    fun initializer(): ProtocolBindingInitializer<TController>
```
Nashatyrev: @raulk I've returned back to `ChannelInitializer` since with `activate(Channel)` I can do only `Pipeline.addLast()` while it's sometimes not desirable since another handlers could (and sometimes should) be installed at the end.

raulk: That's weird. Do you get a runtime exception? Because the API should allow you to do `channel.pipeline().add*()`?

raulk:
Merged this into `sketch`, but this item is still outstanding: https://github.com/raulk/jvm-libp2p-minimal/pull/11#discussion_r300441913.

src/main/kotlin/io/libp2p/core/protocol/ProtocolBinding.kt
Nashatyrev: This might not be exactly our use case, but how I see it: 
You may want in theory to initialize the channel with two handlers: `Secio` + `ApplicationHandler`. Secio handler itself adds and removes his own handlers. If it would do `addLast` then added handlers will be after the `ApplicationHandler` and this will break 'incapulation' of `Secio` handler. 
Thus instead of doing `addLast` is should either `replace` old handlers or call `addBefore/After` its own handlers. 
To achieve this the handler should know 'himself'. This is not possible with `activate(Channel)` approach.

Nashatyrev: That would be ideally handled with kind of 'pipelines hierarchy' but this is not implemented in Netty

PR #15, opened 2019-07-05
Addressing Lint issues

snazha-blkio:


Nashatyrev:
@snazha-blkio So the gradle build completes ok now?

snazha-blkio:
> @snazha-blkio So the gradle build completes ok now?

I don't get any more complaints on this.

PR #20, opened 2019-07-09
Tcp handles

yusefnapora:
I had a hard time falling asleep last night, so I decided to poke around at jvm-libp2p and find something simple to do 😄 

This implements `TcpTransport#handles` and adds a test for it.


Nashatyrev:
Cool! Thanks for PR 👍

PR #21, opened 2019-07-11
Feature/multistream handler

### LINKED ISSUE #17: mplex: implement Netty channel splitter
Implement Netty handler which will be routing stream multiplexer (mplex or other) frames to individual netty channels
Relates to #16

Nashatyrev:
Implemented `MultiplexHandler` which creates and manages child `MultiplexChannel`s
Resolve #17


Nashatyrev:
It seems we need to agree on those class naming. It looks a bit weird now. 
I would first suggest to rename:
(`io.libp2p.core.util.netty.multiplex.`) `MultiplexHandler/MultiplexChannel` => `AbstractMplexHandler/AbstractMplexChannel`

PR #31, opened 2019-07-18
Feature/18 setup logger

snazha-blkio:
Using log4j2.


src/test/kotlin/io/libp2p/core/security/secio/SecIoSecureChannelTest.kt
```diff
@@ -69,6 +70,10 @@ class SecIoSecureChannelTest {
         Assertions.assertEquals("Hello World from 1", rec2)
         Assertions.assertEquals("Hello World from 2", rec1)
     }
+
+    companion object {
+        private val logger = LogManager.getLogger(SecIoSecureChannelTest::class.java)
+    }
```
Nashatyrev: I'd better declare it in one line, like: `private const val logger = `

snazha-blkio: Can't do `const` on non-primitives.

But are you ok to keep them in the companion objects so that they remain as singletons?

snazha-blkio:
@Nashatyrev  thoughts on the `companion object`?

Nashatyrev:
Ok, let's stick with companion object for now. Just hate spending 3 lines instead of 1 for this :)))

Issue #35, opened 2019-07-25
Implement Noise security protocol


Nashatyrev:
https://noiseprotocol.org/noise.html

The Noise usage for Eth2.0 is still under discussion (https://github.com/ethereum/eth2.0-specs/pull/1281/files#r302194647)

Nashatyrev:
https://github.com/libp2p/specs/issues/195

shahankhatch:
branch is here: libp2p/jvm-libp2p/tree/feature/makesomenoise

Nashatyrev:
Done

PR #37, opened 2019-08-02
Feature/makesomenoise

shahankhatch:
PR for #35

### DIFF
--- a/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
+++ b/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
@@ -102,6 +102,7 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
                     }
                 }
                 ctx.fireUserEventTriggered(evt)
+                ctx.fireChannelActive()
             }
         }
         ch.nettyChannel.pipeline().addLast(handshakeHandlerName + chid, NoiseIoHandshake())
--- a/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
+++ b/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
@@ -25,6 +25,7 @@ import io.netty.channel.ChannelInboundHandlerAdapter
 import io.netty.channel.SimpleChannelInboundHandler
 import org.apache.logging.log4j.Level
 import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
 import org.apache.logging.log4j.core.config.Configurator
 import spipe.pb.Spipe
 import java.util.Arrays
@@ -32,42 +33,38 @@ import java.util.concurrent.CompletableFuture
 import java.util.concurrent.atomic.AtomicInteger
 
 class NoiseXXSecureChannel(private val localKey: PrivKey) :
-    SecureChannel {
+        SecureChannel {
 
-    private val logger = LogManager.getLogger(NoiseXXSecureChannel::class.java.name)
+    companion object {
+        const val protocolName = "Noise_XX_25519_ChaChaPoly_SHA256"
+        const val announce = "/noise/$protocolName/0.1.0"
+
+        @JvmStatic
+        var localStaticPrivateKey25519: ByteArray = ByteArray(32)
+
+        init {
+            // initialize static Noise key
+            Noise.random(localStaticPrivateKey25519)
+        }
+    }
+
+    private var logger: Logger
+    private var loggerNameParent: String = NoiseXXSecureChannel::class.java.name + this.hashCode()
+    private lateinit var chid: String
 
     private lateinit var role: AtomicInteger
-    private lateinit var localNoiseState: DHState
-    private var sentNoiseKeyPayload = false
-
-    private val instancePayload = ByteArray(65535)
-    private var instancePayloadLength = 0
 
     private val handshakeHandlerName = "NoiseHandshake"
 
     override val announce = Companion.announce
     override val matcher = ProtocolMatcher(Mode.PREFIX, name = "/noise/$protocolName/0.1.0")
 
-    companion object {
-        const val protocolName = "Noise_XX_25519_ChaChaPoly_SHA256"
-        const val announce = "/noise/$protocolName/0.1.0"
-        @JvmStatic
-        var privateKey25519: ByteArray = ByteArray(32)
-        private var privateKey25519Initialized: Boolean = false
-    }
-
     init {
-        Configurator.setLevel(NoiseXXSecureChannel::class.java.name, Level.DEBUG)
-        synchronized(privateKey25519) {
-            if (!privateKey25519Initialized) {
-                Noise.random(privateKey25519)
-                synchronized(privateKey25519Initialized) {
-                    privateKey25519Initialized = true
-                }
-            }
-        }
+        logger = LogManager.getLogger(loggerNameParent)
+        Configurator.setLevel(loggerNameParent, Level.DEBUG)
     }
 
+    // simplified constructor
     fun initChannel(ch: P2PAbstractChannel): CompletableFuture<SecureChannel.Session> {
         return initChannel(ch, "")
     }
@@ -78,10 +75,8 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
     ): CompletableFuture<SecureChannel.Session> {
         role = if (ch.isInitiator) AtomicInteger(HandshakeState.INITIATOR) else AtomicInteger(HandshakeState.RESPONDER)
 
-        // configure the localDHState with the private
-        // which will automatically generate the corresponding public key
-        localNoiseState = Noise.createDH("25519")
-        localNoiseState.setPrivateKey(privateKey25519, 0)
+        chid = "ch:" + ch.nettyChannel.id().asShortText() + "-" + ch.nettyChannel.localAddress() + "-" + ch.nettyChannel.remoteAddress()
+        logger.debug(chid)
 
         val ret = CompletableFuture<SecureChannel.Session>()
         val resultHandler = object : ChannelInboundHandlerAdapter() {
@@ -99,7 +94,7 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
                     }
                     is SecureChannelFailed -> {
                         ret.completeExceptionally(evt.exception)
-                        ctx.pipeline().remove(handshakeHandlerName)
+                        ctx.pipeline().remove(handshakeHandlerName + chid)
                         ctx.pipeline().remove(this)
                         logger.debug("Reporting secure channel failed")
                     }
@@ -107,28 +102,51 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
                 ctx.fireUserEventTriggered(evt)
             }
         }
-        ch.nettyChannel.pipeline().addLast(handshakeHandlerName, NoiseIoHandshake())
-        ch.nettyChannel.pipeline().addLast(handshakeHandlerName + "ResultHandler", resultHandler)
+        ch.nettyChannel.pipeline().addLast(handshakeHandlerName + chid, NoiseIoHandshake())
+        ch.nettyChannel.pipeline().addLast(handshakeHandlerName + chid + "ResultHandler", resultHandler)
         return ret
     }
 
-    inner class NoiseIoHandshake : SimpleChannelInboundHandler<ByteBuf>() {
+    inner class NoiseIoHandshake() : SimpleChannelInboundHandler<ByteBuf>() {
+
         private val handshakestate: HandshakeState = HandshakeState(protocolName, role.get())
+        private var loggerName: String
+        private var logger2: Logger
+
+        private var localNoiseState: DHState
+        private var sentNoiseKeyPayload = false
+
+        private val instancePayload = ByteArray(65535)
+        private var instancePayloadLength = 0
 
         init {
+            val roleString = if (role.get() == 1) "INIT" else "RESP"
+//            loggerName = loggerNameParent + "." + this.hashCode().toString().substring(5)
+            loggerName = "|" + roleString + "|" + chid + "." + loggerNameParent
+            loggerName = loggerName.replace(".", "_")
+//            System.out.println("loggerName:"+loggerName)
+
+            logger2 = LogManager.getLogger(loggerName)
+            Configurator.setLevel(loggerName, Level.DEBUG)
+
+            logger2.debug("Starting handshake")
+
+            // configure the localDHState with the private
+            // which will automatically generate the corresponding public key
+            localNoiseState = Noise.createDH("25519")
+            localNoiseState.setPrivateKey(localStaticPrivateKey25519, 0)
             handshakestate.localKeyPair.copyFrom(localNoiseState)
             handshakestate.start()
-            logger.debug("Starting handshake")
         }
 
         override fun channelRead0(ctx: ChannelHandlerContext, msg1: ByteBuf) {
-            logger.debug("Starting channelRead0")
+            logger2.debug("Starting channelRead0")
             val msg = msg1.toByteArray()
 
             channelActive(ctx)
 
             if (role.get() == HandshakeState.RESPONDER && flagRemoteVerified && !flagRemoteVerifiedPassed) {
-                logger.error("Responder verification of Remote peer id has failed")
+                logger2.error("Responder verification of Remote peer id has failed")
                 throw Exception("Responder verification of Remote peer id has failed")
             }
 
@@ -156,7 +174,7 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
 
             // after reading messages and setting up state, write next message onto the wire
             if (role.get() == HandshakeState.RESPONDER && handshakestate.action == HandshakeState.WRITE_MESSAGE) {
-                logger.debug("Sending responder noise key payload")
+                logger2.debug("Sending responder noise key payload")
                 sendHandshakePayload(ctx)
             } else if (handshakestate.action == HandshakeState.WRITE_MESSAGE) {
                 val sndmessage = ByteArray(65535)
@@ -169,21 +187,21 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
                 cipherStatePair = handshakestate.split()
                 aliceSplit = cipherStatePair.sender
                 bobSplit = cipherStatePair.receiver
-                logger.debug("Split complete")
+                logger2.debug("Split complete")
 
                 // put alice and bob security sessions into the context and trigger the next action
                 val secureChannelInitialized = SecureChannelInitialized(
-                    NoiseSecureChannelSession(
-                        PeerId.fromPubKey(localKey.publicKey()),
-                        PeerId.random(),
-                        localKey.publicKey(),
-                        aliceSplit,
-                        bobSplit
-                    ) as SecureChannel.Session
+                        NoiseSecureChannelSession(
+                                PeerId.fromPubKey(localKey.publicKey()),
+                                PeerId.random(),
+                                localKey.publicKey(),
+                                aliceSplit,
+                                bobSplit
+                        ) as SecureChannel.Session
                 )
                 ctx.fireUserEventTriggered(secureChannelInitialized)
                 ctx.fireChannelActive()
-                ctx.channel().pipeline().remove(handshakeHandlerName)
+                ctx.channel().pipeline().remove(this)
                 return
             }
         }
@@ -194,14 +212,13 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
             }
             activated = true
 
-            logger.debug("Noise registration starting")
             // even though both the alice and bob parties can have the payload ready
-            // the Noise protocol only permits the alice to send a packet first
+            // the Noise protocol only permits alice to send a packet first
             if (role.get() == HandshakeState.INITIATOR) {
-                logger.debug("Sending initiator noise key payload")
+                logger2.debug("Sending initiator noise key payload")
                 sendHandshakePayload(ctx)
             }
-            logger.debug("Noise registration complete")
+            logger2.debug("Noise registration complete")
         }
 
         /**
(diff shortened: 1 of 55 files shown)


shahankhatch:
Referencing the spec PR that this implementation follows closely.
https://github.com/libp2p/specs/pull/202

src/main/kotlin/tech/pegasys/libp2p/noiseintegration/NoiseOverTcp.kt
```diff
@@ -0,0 +1,34 @@
+package tech.pegasys.libp2p.noiseintegration
```
Nashatyrev: M.b. move this file to `tests` ?

shahankhatch: I've removed this file and `NoiseOverTcpTest.kt` since the test that uses Noise with embedded channels is in `NoiseSecureChannelTest`, which is in `test`.

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+    inner class NoiseIoHandshake : SimpleChannelInboundHandler<ByteBuf>() {
+        private val handshakestate: HandshakeState = HandshakeState(protocolName, role)
+
+        init {
+            handshakestate.localKeyPair.copyFrom(localDHState)
+            handshakestate.start()
+            logger.debug("Starting handshake")
+        }
+
+        override fun channelRead0(ctx: ChannelHandlerContext?, msg: ByteBuf?) {
```
Nashatyrev: When you extending `SimpleChannelInboundHandler` all the read handling should be made within this method and only `ByteBuf` instances are the messages (no `byte[]` are expected here)
Nor `super` call neither `channelRead` calls are expected here.

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        private var activated = false
+        private var flagRemoteVerified = false
+        private var flagRemoteVerifiedPassed = false
+        private var aliceSplit: CipherState? = null
+        private var bobSplit: CipherState? = null
+        private var cipherStatePair: CipherStatePair? = null
+
+        override fun channelRead(ctx: ChannelHandlerContext, msg1: Any) {
```
Nashatyrev: See above comment

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+            logger.debug("Starting channelRead0")
+            val msg = if (msg1 is ByteArray) {
+                msg1
+            } else {
+                (msg1 as ByteBuf).array()
+            }
```
Nashatyrev: Only `ByteBuf` messages are expected in this handler.

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.ChannelInboundHandlerAdapter
+import io.netty.channel.SimpleChannelInboundHandler
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.config.Configurator
+import spipe.pb.Spipe
+import java.util.concurrent.CompletableFuture
+
+class NoiseXXSecureChannel(val localKey: PrivKey, val localDHState: DHState, val remoteDHState: DHState, val role: Int) :
```
Nashatyrev: We are expecting the only `PrivateKey` parameter passed for `SecureChannel`  according to the present API design (check out this `Host` construction sample: https://github.com/libp2p/jvm-libp2p/blob/7585e3e1560da26912f29cc142174c698ed6c6f3/src/test/kotlin/io/libp2p/core/HostTest.kt#L31)

- The `role` here can be taken from `P2PAbstractChannel.isInitiator`
- `remoteDHState` here is never used
- Couldn't the `localDHState` be derived from `Curve25519PrivateKey` ? If yes, then the only thing left to be passed here is the `Curve25519PrivateKey` instance

If the Noise requires something more than just a private key then we need to reconsider design

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+                    // throwing exception for early exit of protocol and for application to handle
+                }
+            }
+
+            // after reading messages and setting up state, write next message onto the wire
+            if (handshakestate.action == HandshakeState.WRITE_MESSAGE) {
+                val sndmessage = ByteArray(65535)
+                val sndmessageLength: Int
+                sndmessageLength = handshakestate.writeMessage(sndmessage, 0, null, 0, 0)
+                ctx.writeAndFlush(sndmessage.copyOfRange(0, sndmessageLength))
```
Nashatyrev: You need to convert message to `ByteBuf` here (`ByteArray.toByteBuf`)

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+                val bs = Spipe.Exchange.newBuilder().setEpubkey(ByteString.copyFrom(localKey.publicKey().bytes()))
+                    .setSignature(ByteString.copyFrom(signed)).build()
+
+                // create the message
+                // also create assign the signed payload
+                val msgLength = handshakestate.writeMessage(msgBuffer, 0, bs.toByteArray(), 0, bs.toByteArray().size)
+
+                // put the message frame which also contains the payload onto the wire
+                val writeAndFlush = ctx?.writeAndFlush(msgBuffer.copyOfRange(0, msgLength))
+                writeAndFlush?.await()
```
Nashatyrev: No blocking calls are allowed within Netty callbacks

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,207 @@
...
+                // generate an appropriate protobuf element
+                val bs = Spipe.Exchange.newBuilder().setEpubkey(ByteString.copyFrom(localKey.publicKey().bytes()))
+                    .setSignature(ByteString.copyFrom(signed)).build()
+
+                // create the message
+                // also create assign the signed payload
+                val msgLength = handshakestate.writeMessage(msgBuffer, 0, bs.toByteArray(), 0, bs.toByteArray().size)
+
+                // put the message frame which also contains the payload onto the wire
+                val writeAndFlush = ctx?.writeAndFlush(msgBuffer.copyOfRange(0, msgLength))
```
Nashatyrev: Arrays should be converted to `ByteBuf`

shahankhatch: done

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Good points about the parameters that can be passed in, with required data generated from the keys. Grabbed the role from P2PAbstractChannel. Done.

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/core/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -83,63 +94,10 @@ class NoiseXXSecureChannel(val localKey: PrivKey, val localDHState: DHState, val
...
-
-        private var activated = false
-        private var flagRemoteVerified = false
-        private var flagRemoteVerifiedPassed = false
-        private var aliceSplit: CipherState? = null
-        private var bobSplit: CipherState? = null
-        private var cipherStatePair: CipherStatePair? = null
-
-        override fun channelRead(ctx: ChannelHandlerContext, msg1: Any) {
+        override fun channelRead0(ctx1: ChannelHandlerContext?, msg1: ByteBuf?) {
```
Nashatyrev: You can safely use non-null types for params here : 
```kt
channelRead0(ctx: ChannelHandlerContext, msg: ByteBuf)
```

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -25,9 +27,11 @@ import org.apache.logging.log4j.core.config.Configurator
 import spipe.pb.Spipe
 import java.util.concurrent.CompletableFuture
 
-class NoiseXXSecureChannel(val localKey: PrivKey, val localDHState: DHState, val remoteDHState: DHState, val role: Int) :
+open class NoiseXXSecureChannel(val localKey: PrivKey, val privateKey25519: ByteArray) :
```
Nashatyrev: Does Noise really require two private keys?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -25,9 +27,11 @@ import org.apache.logging.log4j.core.config.Configurator
...
     SecureChannel {
-    private val logger = LogManager.getLogger(NoiseXXSecureChannel::class.java.name + ":" + role)
+    private var role: Int = HandshakeState.RESPONDER
```
Nashatyrev: Can we make it `lateinit` as well, cause referring this property before `initChannel` is invalid situation

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -202,7 +160,50 @@ class NoiseXXSecureChannel(val localKey: PrivKey, val localDHState: DHState, val
                 ctx.fireUserEventTriggered(secureChannelInitialized)
                 return
             }
-            super.channelRead(ctx, msg1)
         }
+
+        override fun channelRegistered(ctx: ChannelHandlerContext?) {
```
Nashatyrev: You can safely remove `?` from param type

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,206 @@
...
+
+        val ret = CompletableFuture<SecureChannel.Session>()
+        val resultHandler = object : ChannelInboundHandlerAdapter() {
+            override fun userEventTriggered(ctx: ChannelHandlerContext, evt: Any) {
+                when (evt) {
+                    is SecureChannelInitialized -> {
+                        ctx.channel().attr(SECURE_SESSION).set(evt.session)
+                        ret.complete(evt.session)
+                        ctx.pipeline().remove(handshakeHandlerName)
+                        ctx.pipeline().remove(this)
```
Nashatyrev: Sorry, I missed where the messages ciphering happens ?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: One private key is used to verify the peer identity, the other private key is the Noise 25519 key.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Good catch. Since lateinit doesn't work with primitive types, I'm changing it to AtomicInteger for the time being.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: At the moment, it's taking place outside of this class in a wrapping handler. I'll move it into this class.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
Nashatyrev: Trying to dig into this thread now https://github.com/libp2p/specs/issues/195
Yep, the question doesn't look straightforward 

I believe your Telegram channel comment was closely related

> A few points from my end from my very quick perusal of that thread. Take what I say with a grain of salt.
> - As far as I can tell, but maybe I just couldn't find it or didn't recognize some API call, rust-libp2p's current implementation doesn't incorporate any ephemeral keys. This doesn't match Noise protocols like IN, IK, IX (pointed out as not supported on the thread), nor XX (mentioned to be supported in that thread). i.e., I didn't see any DH actions based on ephemeral keys. 
> - As well, as far as I know, transmitting packets requires the use of chaining keys as communication continues, which I couldn't find (maybe I missed it) in the rust-libp2p implementation. 
> - reusing Curve25519 keys in ed25519 or vice-versa is not an easy option. The byte value can be used as an origin private key, but the public keys will be different and will be used for different purposes (ed25519 is for signing and Curve25519 is not?). A (minor) side-effect is that the key interface in jvm-libp2p (which has interface methods for signing/verifying using private/public keys) are not to be implemented.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+
+                        logger.debug("Reporting secure channel failed")
+                    }
+                }
+                ctx.fireUserEventTriggered(evt)
+                ctx.fireChannelActive()
+            }
+        }
+        ch.nettyChannel.pipeline().addLast(handshakeHandlerName + chid, NoiseIoHandshake())
```
Nashatyrev: Why do we need `chid` suffix for the handler name?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+    private lateinit var role: AtomicInteger
+
+    private val handshakeHandlerName = "NoiseHandshake"
+
+    override val announce = Companion.announce
+    override val matcher = ProtocolMatcher(Mode.PREFIX, name = "/noise/$protocolName/0.1.0")
+
+    init {
+        logger = LogManager.getLogger(loggerNameParent)
+        Configurator.setLevel(loggerNameParent, Level.DEBUG)
```
Nashatyrev: I would not use programmatic logging configuration unless it's unavoidable. 
Why `DEBUG` level is hardcoded here?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+        var localStaticPrivateKey25519: ByteArray = ByteArray(32)
+
+        init {
+            // initialize static Noise key
+            Noise.random(localStaticPrivateKey25519)
+        }
+    }
+
+    private var logger: Logger
+    private var loggerNameParent: String = NoiseXXSecureChannel::class.java.name + this.hashCode()
```
Nashatyrev: I'm not sure if creating a logger per instance is a common practice, but anyways I would at least separate it with `.` (like `NoiseXXSecureChannel::class.java.name + '.' + this.hashCode()`) so you may configure the log level in config. 
(also relates to the comment below)

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+    inner class NoiseIoHandshake() : SimpleChannelInboundHandler<ByteBuf>() {
+
+        private val handshakestate: HandshakeState = HandshakeState(protocolName, role.get())
+        private var loggerName: String
+        private var logger2: Logger
+
+        private var localNoiseState: DHState
+        private var sentNoiseKeyPayload = false
+
+        private val instancePayload = ByteArray(65535)
```
Nashatyrev: Can we avoid creating max-length buffer here?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+
+        init {
+            val roleString = if (role.get() == HandshakeState.INITIATOR) "INIT" else "RESP"
+//            loggerName = loggerNameParent + "." + this.hashCode().toString().substring(5)
+            loggerName = roleString + "|" + chid + "|" + loggerNameParent
+            loggerName = loggerName.replace(".", "_")
+//            System.out.println("loggerName:"+loggerName)
+
+            logger2 = LogManager.getLogger(loggerName)
+            Configurator.setLevel(loggerName, Level.DEBUG)
```
Nashatyrev: Why do we need these logger tweaks here? Also the programmatic level (hardcoded) configuration isn't good

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+
+            if (role.get() == HandshakeState.RESPONDER && flagRemoteVerified && !flagRemoteVerifiedPassed) {
+                logger2.error("Responder verification of Remote peer id has failed")
+                ctx.fireUserEventTriggered(SecureChannelFailed(Exception("Responder verification of Remote peer id has failed")))
+                return
+            }
+
+            // we always read from the wire when it's the next action to take
+            // capture any payloads
+            val payload = ByteArray(65535)
```
Nashatyrev: Can we avoid creating max-length buffer here and the similar cases below?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Good points. When I was using `.` as the separator is that only the last segment is displayed in the output, and I'm tracing the instance hierarchy. I also used a logger per instance since log output shows the logger instance name as a separate field in its output and so I figured I would reuse that part, instead of having the instance id appear in the message segment. I'll follow your suggestion and use the `.` separator but keep the instance hierarchy as part of the instance's logger name.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Avoiding max-length buffers requires dynamic bytearrays since the size isn't known beforehand, which will require modifying the Noise library that's been brought in. I'll take a look how to do that.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Good point, since channels are unique I've removed it.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Done.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Done

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Done. The length is now as tight as the longest buffer needed during the handshake (80 extra bytes in the worst case), and it doesn't limit adding arbitrary payloads at any step.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Done.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
shahankhatch: Done.

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
Nashatyrev: >  When I was using . as the separator is that only the last segment is displayed in the output
I believe you just need to change the logging configuration accordingly

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
Nashatyrev: Thanks! Cause e.g. having 100K simultaneous  connections would else require reserving 6G of useless memory

Nashatyrev:
@shahankhatch thanks for addressing my comments 👍 
I'd also made some minor code cleanups

PR #44, opened 2019-08-31
Final clean up

Nashatyrev:
This is code cleanup PR:
- leave only interfaces and core data classes in the `core` package
- add missing javadocs 
- fix issues found during other impls interop


shahankhatch:
lgtm

PR #46, opened 2019-09-25
Feature/test rework

jezhiggins:
Neatening up TcpTransport tests.

As was, testListenClose actually checked three different things, so I've split that into three separate tests. I've also spun testDialClose out into a couple of tests and added some more. I've also made liberal use of helper functions to set the tests up, so hopefully what's actually being tested is clearer. 

It's looks like a lot of commits, but they're all pretty small.


shahankhatch:
lgtm. The build fails on my machine as well as CI, but the failures are in other tests due to an address binding issue (address already in use).

PR #53, opened 2019-10-16
Consolidation

jezhiggins:
Brings together all the inflight branches, so we can move forward together. 

feature/release0.1.0
feature/makesomenoise
test/js-libp2p-interop
fix/host-impl-connection-handler-logic

In addition there are various lint and warning fixes. There's also work on the unit tests - primarily to give them more expressive names, and to split larger tests into smaller, more precise, ones.


jezhiggins:
Any chance we can move this PR along? It contains all the in progress branches, along with a good chunk of work to improve the unit tests, and some testing against other libp2p implementations. I think it would help us all to get this merged down. 

I know it seems like it's enormous, but most of this work is already well known, and the bulk of the remaining changes is relatively small refactoring work. This isn't any particular risk here.

jezhiggins:
Thanks @raulk - you're welcome :slightly_smiling_face: 

The refactoring work I've done in this PR is, naturally enough, towards the end of the sequence of commits. They're primary, if not entirely, in the test cases and as far as I could I only changed one test fixture in each commit. Will that obviously bumps up the number of commits, it hopefully makes reviewing them more straightforward.

Anything further you need me to do, or clarify, or whatever, please do shout.

jezhiggins:
Once this PR is merged down, I've got a pile of work I can submit to follow, including plaintext "secure" channel, so refactoring/simplfication in the muxer, and in secio & noise, and I'm working on websocket transport at the moment. I appreciate the time you're able to give this.

jezhiggins:
I'm pretty sure the spotty test behaviour was pre-existing and this PR doesn't introduce that. I _believe_ at least some of that is due to a race condition in the TcpTransport close down, and I think I might have actually fixed that during a different bit of work today. That will come through in a subsequent PR.

shahankhatch:
Yes, I agree. My apologies if it seemed like my comment was attributing the spotty test behaviour to your PR. I was just mentioning it as an aside.

Nashatyrev:
@jezhiggins thanks for this PR and especially for standalone go/js interop tests inclusion 👍

PR #58, opened 2019-11-18
Refactor/protocol handler

jezhiggins:
Set of refactorings primarily aiming to pull out some core interfaces, and to make writing protocols easier.  See, in particular, the diffs to Identify and Ping protocols for the impact.  Changes elsewhere are comparatively minor, and mainly involve hoisting methods up to the interface they're needed on.

Connection, P2PChannel, and Stream are defined as interfaces. Renamed P2pAbstractHandler to P2PChannelHandler. Throughout - try and avoid accessing nettyChannel member directly, similarly pass parameters directly where possible, use interfaces rather than implementations.


src/main/kotlin/io/libp2p/core/Stream.kt
```diff
@@ -1,26 +1,69 @@
...
 
     /**
      * @return negotiated protocol
      */
-    fun getProtocol(): CompletableFuture<String> = nettyChannel.attr(PROTOCOL).get()
-}
\ No newline at end of file
+    fun getProtocol(): CompletableFuture<String>
+
+    fun <TMessage> pushHandler(protocolHandler: ProtocolMessageHandler<TMessage>) {
```
Nashatyrev: Why do we deal with `Protobuf` in the `Stream` interface?

jezhiggins: What I'm trying to do in this PR is find the "right" place for methods to sit, to try make sure the interfaces have the operations people need. Given that the library is currently quite small, that might mean that some of the choices I've made will turn out to have been too restricting. In that case, we'll just have to make the appropriate changes as the need arises. To me, that always feels better, than going overly liberal. 

In libp2p terms, for example, Connection and Stream are different concepts, representing different things. That fact that we've implemented them using netty.Channel's isn't particularly important, and doesn't make them the same. Nor is it an argument for exposing the netty.Channel. What do, for example, Channel.parent() or Channel.config() mean in the context of a libp2p Stream? 

I pushed the Protobuf handling down into Stream because it seemed to me the best place for it. Again, that might turn out to be wrong in which case we can change it, but my thinking was that *any* protocol that uses a protobuf message needs to add the four Protobuf handlers before adding their own handler. Those four handlers therefore constitute boilerplate. We could document that - "if you use a protobuf, you need to do this" - but if we can document it then we can program it away. Part of the job of a library, surely, is to reduce boilerplate, so that's what I tried to do. Because of the type-erasure inherent in JVM generics, the implementation involves reflection and walking the type hierarchy but there's no way around that, unfortunately. The benefit, as I see it, is that the implementation of Identify, currently our only protobuf based protocol, is similar to Ping and the test protocols we've defined.

ajsutton:
As a user of jvm-libp2p I do find it very confusing to just be given a Netty `ChannelHandlerContext` without any real understanding of what that actually represents. In particular I spent a few hours diving through code trying to understand what `context.channel().disconnect()` did vs `context.channel().close()` and whether it would close/disconnect the write side of the mux stream, the whole mux stream or the entire peer.

Having the API clearly represent the distinction between `Connection` and `Stream`, plus potentially naming things a bit more clearly, would have made it much easier to understand things.  Having access to more powerful Netty APIs makes sense if it's needed but there's a lot you can do without needing that level of access.

src/main/kotlin/io/libp2p/protocol/ProtocolMessageHandler.kt
```diff
@@ -0,0 +1,15 @@
...
+interface ProtocolMessageHandler<TMessage> {
+    fun onActivated(stream: Stream) = Unit
+    fun onMessage(stream: Stream, msg: TMessage) = Unit
+    fun onClosed(stream: Stream) = Unit
+    fun onException(cause: Throwable?) = Unit
+
+    fun fireMessage(stream: Stream, msg: Any) {
+        @Suppress("UNCHECKED_CAST")
+        onMessage(stream, msg as TMessage)
+    }
```
Nashatyrev: Is this method needed in public interface?

src/main/kotlin/io/libp2p/core/Stream.kt
Nashatyrev: Though the libp2p itself is heavily using protobuf the endpoint protocol is free to use any serialization from an unlimited set of possible options. Or for example it may use the protobuf but with alternative message delimiters. You shouldn't handle any possible option here. 

What do you think on this option:
https://github.com/libp2p/jvm-libp2p/commit/80add1550261c4bd9780e90f2f9003b792f3c8f9

Nashatyrev:
@jezhiggins @ajsutton 
guys, I hope I understand your concerns. 

From my point of view you are advocating for simple interfaces  and easy client start up. 
And I agree that netty looks a bit tricky for newcomers. I also agree that creating 5+ classes in current API for a simple client protocol (like Ping) looks like overkill 

On the other hand I'm advocating for being open and flexible and give a user full access to lower level if he needs. I'm not in favor of initially opening narrow functionality subset and then widen it per user requests

Then may be to achieve both goals the right approach could be :
- leave the core API open and transparent (like retaining access to `netty.Channel` from `P2PChannel`)
- create a set of easy-to-use adapters and facades for rapid development of basic client protocols. E.g. `ProtocolHandler` is a good start.  E.g. another handler https://github.com/libp2p/jvm-libp2p/commit/80add1550261c4bd9780e90f2f9003b792f3c8f9 is suitable as a base for protobuf backed protocols

jezhiggins:
These changes don't remove anything - everything that was there before is still there now. It is a move to start creating exactly the adapter and facades you suggest.

Nashatyrev:
> These changes don't remove anything - everything that was there before is still there now.
@jezhiggins I meant to leave `netty.Channel` access in `P2PChannel` but hide the complexity of working with it behind user-friendly 'ProtocolHandler'. 
If I correctly understand your intentions a 'newcomer' user shouldn't even see the `Stream` instance.

src/main/kotlin/io/libp2p/protocol/ProtocolMessageHandler.kt
jezhiggins: It does, because it needs to be accessible from ProtocolMessageHandlerAdapter. Ideally it would be internal, but unfortunately that's not available on interface methods.

jezhiggins:
Anton, now you've cut the 0.2.0 release, can we merge this PR? 

I appreciate you're not entirely happy with it, but I'd like to merge this in so I can then rebase and merge the other work I've done. We can revisit and resolve this in subsequent PRs.

Nashatyrev:
@jezhiggins could you please resolve this concern first: https://github.com/libp2p/jvm-libp2p/pull/58#discussion_r351231474

Nashatyrev:
I would also love to see the `netty.Channel` accessible in `P2PChannel` back unless we found a better approach to hide it

src/main/kotlin/io/libp2p/core/Stream.kt
jezhiggins: It still makes a special case of protobuf messages and requires users to know that they're a special case. With this change I can still write 

```
class IdentifyProtocol(var idMessage: IdentifyOuterClass.Identify? = null) :
    ProtocolHandler<IdentifyController>() {
     ...
}
```

it will compile and, up to a point, it'll even run. But then it'll throw an exception and it's not clear why. I understand your point about not want to accomodate everything in the Stream class, but that isn't what's happening here. We know, because we already have implementations or they're described in libp2p specs, that some protocols are defined using protobufs. It makes sense to me that would should handle those protobuf-using protocols in the lowest friction way possible. That's what I'm trying to do.

jezhiggins:
I'm not sure exposing the `netty.Channel` makes a great deal of sense, particularly in the case of a Stream. As I asked above what do, for example, `Channel.parent()` or `Channel.config()` mean in the context of a libp2p Stream? 

This is why I suggested merging this PR now, so we don't hold up other work which I assume (or at least hope) is less controversial, and we can discuss the details of this separately.

Nashatyrev:
@jezhiggins well, let's leave `Channel` under cover for now.

Just please fix https://github.com/libp2p/jvm-libp2p/pull/58#discussion_r351231474. Since that's the total violation of any OOP principles

jezhiggins:
I disagree. I don't understand how it improves the code. 

Nonetheless, I have applied your change. 

I have pulled the remote branch into this repository and rebased against develop and raised a replacement PR #71.

PR #69, opened 2019-12-01
Release/0.2.0

Nashatyrev:
- Added Noise support according to libp2p Noise spec (initial version) (#53)
- Added Pubsub client validator support (#57)
- Fixed some bugs


src/test/kotlin/io/libp2p/spike/Spike.kt.ignore
```diff
@@ -0,0 +1,85 @@
+package io.libp2p.spike
```
shahankhatch: I think this file's been included by accident?

Nashatyrev: Thanks, removed!

PR #67, opened 2019-11-26
Bugfix/remove log4j config

jezhiggins:
Single change https://github.com/libp2p/jvm-libp2p/commit/69b462b75ce9e9a517a03b0edb8b84bae32100dc to address issue #54

PR #78, opened 2019-12-05
Bugfix/remove log4j config

### LINKED ISSUE #54: Remove log4j2.xml from distributed jar
Currently jvm-libp2p includes a `log4j2.xml` config in `src/main/resources/log4j2.xml` which is then bundled into the production jar. Depending on the class path order, this config from jvm-libp2p may be picked up instead of the application configuration.

Since jvm-libp2p is intended to be used as a library by other applications, the logging configuration should be left to the embedding application, so no log4j config file should be included in the jar.

jezhiggins:
Remove log4j from src/main. 

Not appropriate in a library - users should provide their own log configuration. Closes #54.

Replaces draft PR #67. I've pulled the remote branch into the local repository and rebased on `develop`. It is otherwise identical.


jezhiggins:
Merging as draft PR (okayed by Shahan)[https://github.com/libp2p/jvm-libp2p/pull/67#issuecomment-559201583].

Issue #70, opened 2019-12-04
Support dns4 multiaddrs


ajsutton:
The Prysm Eth2 testnet uses a dns4 Multiaddr for it's bootstrap node which isn't currently supported by jvm-libp2p.

Multiaddr: `/dns4/prylabs.net/tcp/30001/p2p/16Uiu2HAm7Qwe19vz9WzD2Mxn7fXd1vgHHp4iccuyq7TxwRXoAGfc`

Error:
```
io.libp2p.core.Libp2pException: Missing IP4/IP6/DNSADDR in multiaddress /dns4/prylabs.net/tcp/30001/p2p/16Uiu2HAm7Qwe19vz9WzD2Mxn7fXd1vgHHp4iccuyq7TxwRXoAGfc

	at io.libp2p.transport.tcp.TcpTransport.fromMultiaddr(TcpTransport.kt:165)
	at io.libp2p.transport.tcp.TcpTransport.dial(TcpTransport.kt:129)
	at io.libp2p.network.NetworkImpl.connect(NetworkImpl.kt:77)
	at io.libp2p.core.Network$DefaultImpls.connect(Network.kt:55)
	at io.libp2p.network.NetworkImpl.connect(NetworkImpl.kt:15)
...
```

Would be good if jvm-libp2p supported this.

jezhiggins:
I'll see if I can sort this out in the next couple of days.

jezhiggins:
Going well - see feature/dns-resolver branch. Hopefully have it wrapped up sometime tomorrow.

PR #80, opened 2019-12-06
Feature/dns resolver

jezhiggins:
Implements /dns4/hostname and /dns6/hostname resolution, resolving issue #70

### DIFF
--- a/src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
+++ b/src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
@@ -55,12 +55,7 @@ class Multiaddr(val components: List<Pair<Protocol, ByteArray>>) {
      * Queries the address to confirm if it contains the given protocol
      */
     fun has(proto: Protocol): Boolean = getComponent(proto) != null
-    fun hasAny(vararg protos: Protocol): Boolean {
-        for (proto in protos)
-            if (has(proto))
-                return true
-        return false
-    }
+    fun hasAny(vararg protos: Protocol) = protos.any { has(it) }
 
     /**
      * Returns [components] in a human readable form where each protocol value
@@ -205,9 +200,7 @@ class Multiaddr(val components: List<Pair<Protocol, ByteArray>>) {
         }
 
         private fun concatProtocols(parentAddr: Multiaddr, childAddr: Multiaddr): List<Pair<Protocol, ByteArray>> {
-            val protocols = parentAddr.components.toMutableList()
-            protocols.addAll(childAddr.components)
-            return protocols
+            return parentAddr.components + childAddr.components
         }
 
         private fun concatPeerId(addr: Multiaddr, peerId: PeerId): List<Pair<Protocol, ByteArray>> {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/io/libp2p/core/multiformats/MultiaddrDns.kt
```diff
@@ -0,0 +1,116 @@
...
+class MultiaddrDns {
+    interface Resolver {
+        fun resolveDns4(hostname: String): List<Multiaddr>
+        fun resolveDns6(hostname: String): List<Multiaddr>
+    }
+
+    companion object {
+        private val dnsProtocols = arrayOf(Protocol.DNS4, Protocol.DNS6, Protocol.DNSADDR)
+
+        fun resolve(addr: Multiaddr, resolver: Resolver = DefaultResolver): List<Multiaddr> {
```
Nashatyrev: I would consider asynchronous invocation here (aka `CompletableFuture`) since the DNS resolution may take a lot of time.
It also may make sense to check what non-blocking DNS resolve solutions exist to make it possible to resolve DNS in parallel without using a lot of threads

src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
```diff
@@ -96,6 +105,36 @@ class Multiaddr(val components: List<Pair<Protocol, ByteArray>>) {
...
+        )
+        return addresses
+    }
+
+    private fun split(
+        accumulated: MutableList<Multiaddr>,
+        remainingComponents: List<Pair<Protocol, ByteArray>>,
+        pred: (Protocol) -> Boolean
+    ) {
+        val splitIndex = remainingComponents.indexOfLast { pred(it.first) }
```
shahankhatch: I suggest making an explicit type so that terms like `first` aren't ambiguous.

src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
```diff
@@ -165,7 +204,13 @@ class Multiaddr(val components: List<Pair<Protocol, ByteArray>>) {
             return ret
         }
 
-        private fun concatProtocols(addr: Multiaddr, peerId: PeerId): List<Pair<Protocol, ByteArray>> {
+        private fun concatProtocols(parentAddr: Multiaddr, childAddr: Multiaddr): List<Pair<Protocol, ByteArray>> {
+            val protocols = parentAddr.components.toMutableList()
+            protocols.addAll(childAddr.components)
+            return protocols
+        }
```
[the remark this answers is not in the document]
jezhiggins: Ah, cool. Thanks.

src/main/kotlin/io/libp2p/core/multiformats/MultiaddrDns.kt
jezhiggins: This is something to think about, yes.

There is further work to do around DNS - the dnsaddr protocol, and also how we expose this as a configuration item (eg through the DSL) - so I'll come back and address that then.

src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
jezhiggins: Wouldn't do any harm, would it? I'll pick that up in a separate PR.

PR #86, opened 2019-12-13
Feature/mdns discovery

jezhiggins:
Adds Peer Discovery using MDNS.

This is a bit of an unconventional PR and I'd welcome comments and thoughts.

MDNS service discovery allows nodes on the same network segment to discover each other. Both go and js libp2p have an implementation, and they look straightforward to use - basically you just turn them and node start finding each other. I have used them as a guide (and where their implementations differ from the spec, followed what they do). Consequently, the interface I've suggested is minimal, but I believe complete.

```
typealias PeerListener = (PeerInfo) -> Unit

interface Discoverer {
    fun start(): CompletableFuture<Void>
    fun stop(): CompletableFuture<Void>
    fun onPeerFound(listener: PeerListener)
}
```

Unfortunately, the state of Java MDNS implementations is not good. There are only a few and all seemed to be unsupported. Most are incomplete (eg offering discovery but not response). Discarding those, and those for which I could not find source, left only JmDNS. It seems to be very complete. In fact, it turned out to be overly complete - it did much more than we need - while at the same time not providing the flexibility we needed to return response over the wire that was expected. 

I tried to work around things, but in the end I needed to get in and modify the code - largely removing the features we didn't need, and then modifying what remained to suit are needs. I also did a little bit of Java modernisation - mainly swapping out Thread and TimerTasks for Executors. Most of the commits in this PR are, then, the result of pulling the JmDNS code into our tree and then gradually discarding what we don't need. I think the end result is ok, but like i say I'm very happy to hear thoughts.  Thanks.


antonydenyer:
The only concern I've got is around the licensing, the https://github.com/jmdns/jmdns appears to be Apache 2 but the file header in the class is LPGL which can be problematic for some organisations.

jezhiggins:
> The only concern I've got is around the licensing, the https://github.com/jmdns/jmdns appears to be Apache 2 but the file header in the class is LPGL which can be problematic for some organisations.

Good spot. I will double check the licensing.

jezhiggins:
The JmDNS code *is* licensed under the Apache 2.0 License. 

The JmDNS repository has [a note](https://github.com/jmdns/jmdns/blob/master/NOTICE.txt) which says the code was originally licensed under LGPL, but was relicensed in 2014 under the Apache license. 

I'll update the JmDMS derived code to make sure it has appropriate attribution.

PR #87, opened 2019-12-16
Fix for failing Noise test due to uninitialized variable.

shahankhatch:


jezhiggins:
NoiseXXSecureChannel.announce should be fully initialised by the time its accessed, and this test has never failed for me. Are you sure?

shahankhatch:
Rerunning the NoiseXX test on the branch without this proposed fix passes without the error I was seeing when I submitted this PR. I'm closing this PR unless there's better evidence that this fix is needed. Thanks for questioning this PR.

PR #88, opened 2020-01-14
Release/0.4.0

jezhiggins:
Adds

- MDNS-SD discovery


src/main/kotlin/io/libp2p/core/Discoverer.kt
```diff
@@ -0,0 +1,11 @@
+package io.libp2p.core
+
+import java.util.concurrent.CompletableFuture
+
+typealias PeerListener = (PeerInfo) -> Unit
+
+interface Discoverer {
+    fun start(): CompletableFuture<Void>
+    fun stop(): CompletableFuture<Void>
+    fun onPeerFound(listener: PeerListener)
```
Nashatyrev: From my perspective it's a bit odd name for the method for adding a listener. 
For me the following looks nice: 
```kt
interface Discoverer {
    fun start(): CompletableFuture<Void>
    fun stop(): CompletableFuture<Void>
    val newPeerFoundListeners: MutableCollection<PeerListener>
}
...
discoverer.newPeerFoundListeners += { println(it) }
```
However this is a matter of taste

src/main/kotlin/io/libp2p/discovery/MDnsDiscovery.kt
```diff
@@ -0,0 +1,132 @@
...
+import java.net.Inet6Address
+
+class MDnsDiscovery(
+    private val host: Host,
+    private val serviceTag: String = ServiceTagLocal,
+    private val queryInterval: Int = QueryInterval,
+    val address: InetAddress? = null
+) : Discoverer {
+    private val localhost = InetAddress.getLocalHost()
+    private var mDns = JmDNS.create(address ?: localhost)
```
Nashatyrev: `val` also suites here

jezhiggins:
@Nashatyrev Just a quick note to say that I haven't forgotten about this and I do appreciate your time. I've been tied up with other things, but will come back to it. What I'll probably do, is withdraw this PR, resolve these things on the develop branch, and the reraise the merge to release.

PR #91, opened 2020-01-22
Fix Stream.isInitiator

Nashatyrev:
The `Stream.isInitiator` is not the same as `Connection.isInitiator`


Nashatyrev:
Wondering why this regression wasn't caught by our tests

PR #96, opened 2020-03-16
Report error with context and full stack trace

### LINKED ISSUE #95: "ERROR | SecIoSecureChannel | null" log message
When connecting to peers on Lighthouse's testnet 5, we get a number of unhelpful errors in the log:
```
2020-03-17 09:35:55.124+10:00 | nioEventLoopGroup-3-19 | ERROR | SecIoSecureChannel | null
```

Most likely a `NullPointerException` is being thrown somewhere. It would be much more helpful if the full stack trace was included with the error, instead of just the exception's message.

ajsutton:
When SecIo handshake fails, include the full stack trace instead of just the message.  Otherwise `NullPointerException` becomes a very unhelpful `null` error message with no context.

fixes #95


ajsutton:
Superseded by https://github.com/libp2p/jvm-libp2p/pull/102

PR #102, opened 2020-05-06
Reduce log levels for expected conditions

ajsutton:
Remote peers disconnecting or sending invalid content is a common occurrence and shouldn't result on error log messages.

Additionally at the reduced log level, ensure that the full exception is made available to the logging system so that the log configuration can choose whether or not to include the full stack trace.


Nashatyrev:
Merged to `0.3.x` with https://github.com/libp2p/jvm-libp2p/pull/104

PR #108, opened 2020-05-14
Implement gossip v1.1

Nashatyrev:
## PR Description

Implementation of Gossip 1.1 protocol: 
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md

Limitation: 
- [Peer Exchange](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange) implemented without 'Signed peer records' (which is optional)

Public API changes: 
- Subscriber validation result can also be `Ignore` when message is valid but shouldn't be propagated
- Much more router and score parameters (totally >50 params) can now be passed to the `GossipRouter`
- Among these parameters there is `GossipParamsV1_1.connectCallback` which is invoked whenever `GossipRouter` wants the external peer manager to honor and connect to specific peers 
- Replace `ProtocolBinding` `announce+matcher` properties with new `ProtocolDescriptor` class (https://github.com/libp2p/jvm-libp2p/pull/108/commits/9b7a65c40fcdf8810a1409f8217c529a21a48661)
- Replace `ProtocolMatcher` (which is now a part of `ProtocolDescriptor`) with a simple predicate interface

## TODO: 
- [x] `1.0` and `1.1` compatibility
- [x] check Go Gossip 1.1 interoperability
- [x] Check and setup reasonable default score parameters 
- [x] Check Java interoperability


cemozerr:
> * [Peer Exchange](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange) implemented without 'Signed peer records' (which is optional)

How do we implement Peer Exchange without Signed peer records? The following line from spec makes me think that PX is only possible when `PRUNE` control messages include the optional set of peers.

Nashatyrev:
> 
> 
> > * [Peer Exchange](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange) implemented without 'Signed peer records' (which is optional)
> 
> How do we implement Peer Exchange without Signed peer records? The following line from spec makes me think that PX is only possible when `PRUNE` control messages include the optional set of peers.

`Signed Peer Records` seemed to be pretty large RFC so I would omit it for this release for the sake of resources economy. 

This version exchanges only `PeerId`s without address information which can also be helpful: if the discovery system knows the peer by id then the connection manager may favor those peers over others or add them additional reputation points or whatever (I'm not sure we have such mechanism at the moment)  

In short I would postpone SPR feature until it's actually wanted for Teku

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
+    private fun emitGossip(topic: Topic, excludePeers: Collection<PeerHandler>) {
+        val ids = mCache.getMessageIds(topic)
+        if (ids.isEmpty()) return
+
+        val shuffledMessageIds = ids.shuffled(random).take(params.maxIHaveLength)
+        val peers = (getTopicPeers(topic) - excludePeers)
+            .filter { score.score(it) >= score.params.gossipThreshold && !isDirect(it) }
+
+        peers.shuffled(random)
+            .take(max((params.gossipFactor * peers.size).toInt(), coreParams.DLazy))
```
cemozerr, line 398: I think this should be:
```suggestion
            .take(min((params.gossipFactor * peers.size).toInt(), coreParams.DLazy))
```
Since the spec says:
> From these peers, it randomly selects gossip factor peers with a minimum of D_lazy, and emits gossip to the selected peers.

cemozerr: And it might be a good idea to add a test for it

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
+                } else if (peers.size > coreParams.DHigh) {
+                    // too many mesh peers
+                    val sortedPeers = peers
+                        .shuffled(random)
+                        .sortedBy { score.score(it) }
+                        .reversed()
+
+                    val bestDPeers = sortedPeers.take(coreParams.DScore)
+                    val restPeers = sortedPeers.drop(coreParams.DScore).shuffled(random)
+                    val outboundCount = bestDPeers.count { it.isOutbound() }
```
cemozerr, line 333: I think this should be:
```suggestion
                    val outboundCount = (bestDPeers + restPeers).count { it.isOutbound() }
```
As the spec says:
> **if the scoring plus random selection** does not result in enough outbound connections, then we replace the random and lower scoring peers in the selection with outboud connection peers.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
-                        prune(dropPeer, topic)
+
+                // drop underscored peers from mesh
+                peers.filter { score.score(it) < 0 }
+                    .forEach { prune(it, topic) }
+
+                if (peers.size < coreParams.DLow) {
+                    // need more mesh peers
+                    (getTopicPeers(topic) - peers)
+                        .filter { score.score(it) >= 0 && !isDirect(it) && !isBackOff(it, topic) }
```
cemozerr, line 320: I know the spec does not say anything about this, but I'm curious if we should, on top of ignoring peers with a negative score, favor peers with a better score.

cemozerr:
I could not find any modifications about Explicit Peering Agreements. I'm assuming it's not necessary for Eth2.0/Teku, so implementing it is not a priority for now?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
-                    (getTopicPeers(topic) - peers).shuffled(random).take(D - peers.size).forEach { newPeer ->
-                        peers += newPeer
-                        graft(newPeer, topic)
-                    }
-                } else if (peers.size > DHigh) {
-                    peers.shuffled(random).take(peers.size - D).forEach { dropPeer ->
-                        peers -= dropPeer
-                        prune(dropPeer, topic)
+
+                // drop underscored peers from mesh
```
cemozerr, line 313: Should we not be doing the same thing to fanout topic and peers?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -65,27 +96,133 @@ open class GossipRouter : AbstractRouter() {
...
+        val iWant = msg.messageIDsList - seenMessages.keys
+        val asked = iAsked.computeIfAbsent(peer) { AtomicInteger() }
+        val maxToAsk = max(0, min(iWant.size, params.maxIHaveLength - asked.get()))
+        asked.addAndGet(iWant.size)
+        iWant(peer, iWant.shuffled(random).subList(0, maxToAsk))
+    }
+
+    private fun handleIWant(msg: Rpc.ControlIWant, peer: PeerHandler) {
+        val peerScore = score.score(peer)
+        if (peerScore < score.params.gossipThreshold) return
```
cemozerr, line 213: is IWant message considered a gossip message as well?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
                 }
-                submitGossip(topic, peers)
+
+                emitGossip(topic, peers)
             }
             fanout.entries.forEach { (topic, peers) ->
-                peers.removeIf { it in getTopicPeers(topic) }
-                val needMore = D - peers.size
+                peers.removeIf {
+                    it !in getTopicPeers(topic) || score.score(it) < scoreParams.publishThreshold
```
cemozerr, line 365: The spec suggests that `publishThreshold` is only used when flood publishing. Is fanout a type of flood publishing? 
> publishThreshold: when a peer's score drops below this threshold, self published messages are not propagated towards this peer when (flood) publishing.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -0,0 +1,213 @@
...
+            this@GossipScore.params.decayToZero,
+            params.MeshMessageDeliveriesCap
+        )
+        var meshFailurePenalty: Double by cappedDouble(0.0, this@GossipScore.params.decayToZero)
+        var invalidMessages: Double by cappedDouble(0.0, this@GossipScore.params.decayToZero)
+
+        fun inMesh() = joinedMeshTime > 0
+
+        fun meshTimeNorm() = min(
+            (if (inMesh()) curTime() - joinedMeshTime else 0).toDouble() / params.TimeInMeshQuantum.toMillis(),
```
cemozerr, line 44: The spec suggests that the _"The time in mesh is calculated lazily during the decay update to avoid a large number of calls to `gettimeofday`_". 

I might not be correctly understanding the lazy keyword concept in Kotlin, but it seems to me that we are not lazily updating `meshTime` during the decay update, but calling `System.currentTimeMillis()` at every `score()` call.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -0,0 +1,213 @@
...
+class GossipScore(
+    val params: GossipScoreParams = GossipScoreParams(),
+    val executor: ScheduledExecutorService,
+    val curTime: () -> Long
+) {
+
+    inner class TopicScores(val topic: Topic) {
+        private val params = topicParams[topic]
+
+        var joinedMeshTime: Long = 0
```
cemozerr, line 27: nit:
```suggestion
        var joinedMeshTimeInMillis: Long = 0
```

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -0,0 +1,213 @@
...
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.pow
+
+class GossipScore(
+    val params: GossipScoreParams = GossipScoreParams(),
+    val executor: ScheduledExecutorService,
+    val curTime: () -> Long
```
cemozerr, line 21: nit:
```suggestion
    val curTimeInMillis: () -> Long
```

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -0,0 +1,213 @@
...
+
+        fun inMesh() = joinedMeshTime > 0
+
+        fun meshTimeNorm() = min(
+            (if (inMesh()) curTime() - joinedMeshTime else 0).toDouble() / params.TimeInMeshQuantum.toMillis(),
+            params.TimeInMeshCap
+        )
+
+        fun isMeshMessageDeliveriesActive() =
+            inMesh() && ((curTime() - joinedMeshTime).millis > params.MeshMessageDeliveriesActivation)
```
cemozerr, line 49: Spec again suggests this `curTime()` to be calculated lazily during the decay update.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -65,27 +96,133 @@ open class GossipRouter : AbstractRouter() {
...
+        }
+
+        val iWant = msg.messageIDsList - seenMessages.keys
+        val asked = iAsked.computeIfAbsent(peer) { AtomicInteger() }
+        val maxToAsk = max(0, min(iWant.size, params.maxIHaveLength - asked.get()))
+        asked.addAndGet(iWant.size)
+        iWant(peer, iWant.shuffled(random).subList(0, maxToAsk))
+    }
+
+    private fun handleIWant(msg: Rpc.ControlIWant, peer: PeerHandler) {
```
cemozerr, line 211: I believe we have to implement this:
> IWANT message responses are limited in the number of retransmissions to a certain peer; in gossipsub v1.0 the router always responds to IWANT messages when the message in the cache. In gossipsub v1.1 the router responds a limited number of times to each peer so that IWANT spam does not cause a signficant drain of resources.

Unless you implemented this somewhere else and I'm missing it.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -65,27 +96,133 @@ open class GossipRouter : AbstractRouter() {
...
+        val peerScore = score.score(peer)
+        // we ignore IHAVE gossip from any peer whose score is below the gossip threshold
+        if (peerScore < score.params.gossipThreshold) return
+        if (peerIHave.computeIfAbsent(peer) { AtomicInteger() }.incrementAndGet() > params.maxIHaveMessages) {
+            return
+        }
+
+        val iWant = msg.messageIDsList - seenMessages.keys
+        val asked = iAsked.computeIfAbsent(peer) { AtomicInteger() }
+        val maxToAsk = max(0, min(iWant.size, params.maxIHaveLength - asked.get()))
```
cemozerr, line 206: Is the number of maximum ids in an `IWANT` equal to the number of maximum ids in an `IHAVE` message?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -98,104 +235,213 @@ open class GossipRouter : AbstractRouter() {
...
+    private fun prune(peer: PeerHandler, topic: Topic) {
+        setBackOff(peer, topic)
+        enqueuePrune(peer, topic)
+        if (mesh[topic]?.remove(peer) == true) {
+            notifyPruned(peer, topic)
+        }
+    }
+
+    private fun iWant(peer: PeerHandler, messageIds: List<MessageId>) {
+        messageIds.forEach { iWantRequests[peer to it] = curTime() }
```
cemozerr, line 417: It seems like the spec is suggesting to track only one random message-id, not all of the message-ids.
> For each IHAVE advertisement which elicits an IWANT request, the router tracks a random message ID within the advertised set.

src/main/kotlin/io/libp2p/core/pubsub/PubsubApi.kt
```diff
@@ -12,11 +12,33 @@ import kotlin.random.Random.Default.nextLong
...
+ */
+enum class ValidationResult {
+
+    /**
+     * The message is valid and should be propagated further
+     */
+    Valid,
+
+    /**
+     * The message is invalid
```
cemozerr, line 26: nit: 
```suggestion
     * The message is invalid, should be rejected, and trigger penalty.
```

src/main/kotlin/io/libp2p/core/pubsub/PubsubApi.kt
```diff
@@ -12,11 +12,33 @@ import kotlin.random.Random.Default.nextLong
...
+     */
+    Invalid,
+
+    /**
+     * The message is valid but should NOT be propagated further
```
cemozerr, line 31: nit:
```suggestion
     * The message should not be propagated further but does not trigger a penalty.
```

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: From my understanding of the spec `D_lazy` is the lower bound so if the `params.gossipFactor * peers.size` value is lower than `D_lazy` then the `D_lazy` value should be used. This is `max`. 
May be I missed something?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev, line 333: @cemozerr seems like you deleted your comment here, but it was near correct
It should be
```kt
val outboundCount = (bestDPeers + restPeers)
    .take(coreParams.D)
    .count { it.isOutbound() }
```

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Do you mean sort the peers by score before taking desired number of peers? 
There is no such requirement and the reference Go implementation doesn't do it: https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1283

I believe there could be some sense behind this: we don't want total score biasing, there should be some randomness. But I hope @vyzo might want to comment on this

Nashatyrev:
> 
> 
> I could not find any modifications about Explicit Peering Agreements. I'm assuming it's not necessary for Eth2.0/Teku, so implementing it is not a priority for now?

We are handling explicit (or direct) peers via this Gossip parameter callback: https://github.com/libp2p/jvm-libp2p/blob/7fc89c886a04fddafc123c3dc14d19592cbd92ba/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt#L80

Yep, the Gossip spec says that the gossip _router_ should establish check and manage connection to explicit peers. However I'm not sure if it makes sense to empower the pubsub router with such connection manager duties. I believe if/when the client application would like to manage such explicit peers

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Fanout peers are not in the mesh and we can't prune them, however we are also cleaning our fanout list from underscored peers: 
https://github.com/libp2p/jvm-libp2p/blob/7fc89c886a04fddafc123c3dc14d19592cbd92ba/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt#L365

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Yep, `IWANT` is a purely Gossip message

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Hmm, good question! 
That condition was taken from the Go implementation: https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1440-L1442
But I also didn't find any mentions in the spec. 
@vyzo may be you can comment?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: The current implementation is very far from the optimal with regards to performance. `System.currentTimeMillis()` is absolutely not the bottleneck at the moment. 
I'm not sure why that is mentioned in the spec. 
May be @vyzo could comment on this?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Good catch! 👍 
Missed that restriction

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: If you don't mind I would change to `joinedMeshTimeMillis`

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: If you don't mind I would change to `curTimeMillis`

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: Commented above

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Right, that makes sense! Else the penalty can be too large for a single missed `IWANT`

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
cemozerr: Yes, you're right. Sorry for the confusion.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
vyzo: Yes indeed, we want random selection!

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
vyzo: Yeah, the publish threshold applies everywhere there is an actual publication.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
vyzo: It's very fast in most OSes, due to VDSO.

vyzo: But then again some OSes have a very slow implementation (a syscall) -- see OpenBSD.

vyzo: It is mentioned in the spec so that there is leeway to implement it this way, if you care to avoid calling `gettimeofday` in every score invocation -- it could be called potentially for every peer within the heartbeat, hence the reason for laziness.

vyzo: Another point that is relevant, with the lazy implementation we can avoid biasing the time in mesh by order of score invocation. This _could_ be an issue if you have a slow `gettimeofday` in your system.

vyzo:
> Yep, the Gossip spec says that the gossip router should establish check and manage connection to explicit peers. However I'm not sure if it makes sense to empower the pubsub router with such connection manager duties. I believe if/when the client application needs such kind of dedicated peers they should be managed on a higher level and the Gossip router should be just notified that those peers are explicit. Again calling for @vyzo to comment on this

This is fine, you can implement it this way if you so desire. However, it should be pointed out that there should be some logic _somewhere_ for maintaining connections for explicit peering agreements. We implemented this way in the reference implementation because it was easy and it was requested as an explicit feature.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Done: https://github.com/libp2p/jvm-libp2p/commit/94a02d0c9dae34d9fdc18ab79a61d99fc9560e49

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Yep, it's not the part of the spec, but here is the Go implementation doc on this param: 
https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L150-L155

Nashatyrev: However I'm not sure about the logic of how this parameter is processed in Go. Seems slightly odd. Will revisit it again

Nashatyrev:
There are still no reasonable average defaults for scoring params, and Go implementation left this onto client code. I.e. the default params mean _no scoring_

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Hm, now looks ok to me. Add a test for this and fix a bug

src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
```diff
@@ -10,62 +10,222 @@ import kotlin.math.min
...
+/**
+ * Parameters of Gossip 1.1 router
+ */
+data class GossipParams(
+    /**
+     * [D] sets the optimal degree for a GossipSub topic mesh. For example, if [D == 6],
+     * each peer will want to have about six peers in their mesh for each topic they're subscribed to.
+     * [D] should be set somewhere between [DLow] and [DHi].
+     */
     val D: Int = 3,
```
cemozerr, line 32: nit: Isn't this a bit low?

Nashatyrev: Right 👍 
`3` is some ancient default. `6` is in the Go impl

PR #110, opened 2020-05-21
0.4.1: Resolve problems with large packets 

Nashatyrev:
There were several problems with sending large packets: 
- Noise frame limit of 65Kb 
- Mplex couldn't correctly handle fragmented frames


src/main/kotlin/io/libp2p/etc/util/netty/SplitEncoder.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Splits large outbound buffers onto smaller chunks of specified max size
+ */
+class SplitEncoder(val maxSize: Int) : MessageToMessageEncoder<ByteBuf>() {
+    override fun encode(ctx: ChannelHandlerContext, msg: ByteBuf, out: MutableList<Any>) {
+        var startIdx = msg.readerIndex()
+        while (startIdx < msg.writerIndex()) {
+            val endIndex = min(startIdx + maxSize, msg.writerIndex())
+            out += msg.retainedSlice(startIdx, endIndex - startIdx)
+            startIdx += maxSize
```
cemozerr, line 17: Should you not be incrementing `startIdx` here by `endIndex - startIndex` ?

src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
```diff
@@ -33,9 +34,11 @@ private enum class Role(val intVal: Int) { INIT(HandshakeState.INITIATOR), RESP(
 
 private val log = LogManager.getLogger(NoiseXXSecureChannel::class.java)
 private const val HandshakeNettyHandlerName = "HandshakeNettyHandler"
+private const val NoiseCodeNettyHandlerName = "NoiseXXCodec"
+private const val MaxCipheredPacketLength = 65535
 
 class UShortLengthCodec : CombinedChannelDuplexHandler<LengthFieldBasedFrameDecoder, LengthFieldPrepender>(
-    LengthFieldBasedFrameDecoder(0xFFFF, 0, 2, 0, 2),
+    LengthFieldBasedFrameDecoder(0xFFFF + 2, 0, 2, 0, 2),
```
mbaxter, line 41: Is the frame length coming from the constant above?:
```suggestion
    LengthFieldBasedFrameDecoder(MaxCipheredPacketLength + 2, 0, 2, 0, 2),
```

src/main/kotlin/io/libp2p/etc/util/netty/SplitEncoder.kt
Nashatyrev: `endIndex - startIndex` would differ from `maxSize` on the terminal iteration only when the loop exit

src/main/kotlin/io/libp2p/mux/mplex/MplexFrameCodec.kt
```diff
@@ -54,8 +51,16 @@ class MplexFrameCodec : MessageToMessageCodec<ByteBuf, MuxFrame>() {
...
     override fun decode(ctx: ChannelHandlerContext, msg: ByteBuf, out: MutableList<Any>) {
         while (msg.isReadable) {
+            val readerIndex = msg.readerIndex()
             val header = msg.readUvarint()
             val lenData = msg.readUvarint()
+            if (header < 0 || lenData < 0 || msg.readableBytes() < lenData) {
+                msg.readerIndex(readerIndex)
+                return
+            }
+            if (lenData > MaxMessageSize) {
```
cemozerr, line 61: Why do you skip to the end of `lenData` or `messageSize`, whichever is smaller, if the `lenData` is greater than `MaxMessageSize`?

cemozerr: Do you basically skip a packet if its `lenData` is greater than `MaxMessageSize`?

src/test/kotlin/io/libp2p/security/SecureChannelTest.kt
```diff
@@ -57,25 +70,36 @@ abstract class SecureChannelTest(
...
+            val trimDataSize = dataSize - 2
+            val data = (0 until trimDataSize / 10)
+                .map { " x" + Integer.toHexString(it * 10).padStart(8, '0') }
+                .joinToString("")
+                .padEnd(trimDataSize, '_')
+            data1 = "1-$data"
+            data2 = "2-$data"
+        } else {
+            data1 = if (dataSize == 1) "1" else ""
+            data2 = if (dataSize == 1) "2" else ""
```
mbaxter, line 85: Doesn't look like `dataSize` is ever `== 1`

src/main/kotlin/io/libp2p/mux/mplex/MplexFrameCodec.kt
Nashatyrev: Good question. I actually didn't know what the best we can do here. Just released the data so it is not accumulated further. 
We should probably at least throw exception and at most close connection since the protocol is violated without any recovery possibility

cemozerr: Yeah, I’d err on the side of throwing an exception here since it’s an unsupported operation and something is clearly wrong. Closing the connection sounds a a bit harsh. That’s me assuming the connection is recoverable. i.e. we miss this one message, but do get their future messages properly. Not sure if that’s the case.

mbaxter: Seems like we should log an error, and close the stream.

Nashatyrev: I don't think we would be able to accept following data.
That most likely means the protocol is broken and we couldn't parse further data
I would just better close the connection right here, since it might be a security breach when trying to handle such situations

src/test/kotlin/io/libp2p/security/SecureChannelTest.kt
Nashatyrev: Fix the condition above. Yep there is no cases with size 1, but just made this piece of code universal :)

src/main/kotlin/io/libp2p/mux/mplex/MplexFrameCodec.kt
Nashatyrev: What I finally did is throwing exception when `lenData > MaxMessageSize`
In the `exceptionCaught`: 
- propagate exception to let the following handlers know what's going on and why we are going to close channel. I would also prefer to keep error logging in a single place (it's now the last handler: `AbstractMuxHandler`) and not spreading it across handlers 
- `close()` the channel since if any error happens in the `decode()` method then is it likely unrecoverable. Any errors in lower level handlers (like cipher or frame length decoder) are also most likely unrecoverable

PR #112, opened 2020-06-04
Release 0.5.1

### LINKED ISSUE #94: Stream handler channelRead method invoked before channelActive
### Issue Description
I've been testing teku's syncing logic and have noticed an issue with the ordering of events for stream handlers.  For reference, our stream-related setup logic is [here](https://github.com/PegaSysEng/artemis/blob/c5748972645aaf4ef937e99419e06c957398d06a/networking/p2p/src/main/java/tech/pegasys/artemis/networking/p2p/libp2p/rpc/RpcHandler.java#L39).

When a remote peer initiates a request stream with our node, we push a `SimpleChannelInboundHandler` to manage these requests and I've noticed that this handler's `channelRead0` method is sometimes invoked before `channelActive`.  From discussing with @Nashatyrev, it sounds like this may be a bug.

Nashatyrev:
## Description

A couple of issues fixed
- (#94): we are now able to handle multistream zero-roundtrip negotiations: when the initiator sends requested protocol id immediately followed by the protocol payload 
- While fixing the initial issue it turned out that `MultistreamImpl` results in the incorrect position of `ProtocolSelect` handler relative to `Negotiator` handler. This is because `Negotiator` is added via `ChannelInitializer` and `addLast` is used during initialization. That way even though the `ProtocolSelect` is added _after_ `Negotiator` initializer it may stay _before_ if the channel is inited asynchronously. This issue was fixed as well

## Fixes
fixes #94
should resolve https://github.com/PegaSysEng/teku/issues/1968
should resolve https://github.com/PegaSysEng/teku/issues/2031


src/main/kotlin/io/libp2p/multistream/ProtocolSelect.kt
```diff
@@ -21,15 +22,38 @@ class ProtocolSelect<TController>(val protocols: List<ProtocolBinding<TControlle
...
+        // See https://github.com/libp2p/jvm-libp2p/issues/94
+        activeFired = true
+        ctx.fireChannelActive()
+        ctx.fireChannelRead(msg)
+    }
+
+    override fun channelActive(ctx: ChannelHandlerContext) {
+        if (!userEvent) {
+            println("Ops")
+        }
```
ajsutton, line 41: Not sure a println should be here?

Nashatyrev: Oh nice catch 👍

PR #118, opened 2020-07-09
Add method to API to retrieve complete list of subscribed topics by peer

ajsutton:
This exposes the `peerTopics` variable via the public API which is useful for clients needing to evaluate how many peers are subscribed to particular topics.

Not entirely sure this is the right way to fit it into the APIs but think it is.  The big downside is that internally topics are tracked as `String` but the `PubsubSubscriberApi` wraps them in a `Topic` type.  That results in having to iterate through the data multiple times - once to copy the current state and once to wrap the `String` in `Topic`.

### DIFF
--- a/src/main/kotlin/io/libp2p/pubsub/PubsubApiImpl.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/PubsubApiImpl.kt
@@ -107,7 +107,7 @@ open class PubsubApiImpl(val router: PubsubRouter) : PubsubApi {
     override fun getPeerTopics(): CompletableFuture<Map<PeerId, Set<Topic>>> {
         return router.getPeerTopics().thenApply { peerTopics ->
             peerTopics.mapValues { topicNames ->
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/libp2p/pubsub/PubsubApiImpl.kt
```diff
@@ -104,6 +104,14 @@ open class PubsubApiImpl(val router: PubsubRouter) : PubsubApi {
         return subscription
     }
 
+    override fun getPeerTopics(): CompletableFuture<Map<PeerId, Set<Topic>>> {
+        return router.getPeerTopics().thenApply { peerTopics ->
+            peerTopics.mapValues { topicNames ->
+                topicNames.value.mapTo(HashSet()) { topicName -> Topic(topicName) } as Set<Topic>
```
Nashatyrev, line 110:
```suggestion
                topicNames.value.mapTo(HashSet()) { topicName -> Topic(topicName) }
```
This cast looks obsolete here

PR #136, opened 2020-08-28
Fix the panic when SemiDuplex has no outbound stream yet

Nashatyrev:
- Fix CompletableFuture anyComplete() implementation
- Tolerate the case when trying writing to a peer without established outbound stream within P2PServiceSemiDuplex

Resolves https://github.com/PegaSysEng/teku/issues/2620


ajsutton (approved):
LGTM.

PR #138, opened 2020-09-04
Do not set omitted fields to default values when regossiping

ajsutton:
`from`, `seqNo` and `signature` are all optional fields.  On parsing they are allowed to be optional, but for `from` and `seqNo` previously the parsed message would set them to default values and when the messages were regossiped, the default values became explicitly set.  This caused the regossiped messages to be rejected by Lighthouse's stricter gossip 1.1 validation.

This does result in a public API change as from and seqId are now optional.  Additionally to enable testing, the seqIdGenerator can now return a null sequence number allowing the generation of messages with no sequence number.


rolfyone (approved):

Nashatyrev (approved):
Good 👍 
That's what I was also thinking about! Thanks

PR #139, opened 2020-09-07
Fix Identify protocol stream close

Nashatyrev:
`Identify` protocol should `CLOSE` the stream after writing the data instead of `RESET`

Fixes https://github.com/PegaSysEng/teku/issues/2728


ajsutton (approved):
LGTM. Confirmed that rumor now picks up the teku user agent correctly.

PR #140, opened 2020-09-07
Upgrade gradle to 6.6.1

ajsutton:
Adds support for JDK 14
Upgrade google protobuf plugin for compatibility with new gradle
Remove buildscan plugin as it is now only available in gradle enterprise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk:
   - openjdk8
 
 script:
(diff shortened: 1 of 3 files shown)


rolfyone (approved):
LGTM

Nashatyrev (approved):
Thanks for handling this 👍

PR #145, opened 2020-10-21
Add seenTTL gossip parameter and its handling

### LINKED ISSUE #130: Implement Gossip `seen_ttl` parameter
Currently the seen message buffer is an LRU Set with hardcoded length `10000`
We need more flexible configuration on this structure (https://github.com/ethereum/eth2.0-specs/pull/1958): 
- **`seen_ttl`**: how many heartbeats the message ID shouldn't be pruned from the buffer
- **`seen_max`**: should limit the buffer size (to avoid out of mem) and in case of overflow delete older entries overriding previous parameter behavior

Nashatyrev:
- Make `AbstractRouter.seenMessages` open but left it defaulting to a simple `LRUMap`
- Override the `seenMessage` Map in `GossipRouter` and made it a `TimeLimitedMap`
- Add a couple of tests 
- Add `seenTTL` param to the `GossipParamsBuilder`

Fix #130 
Fix https://github.com/quantstamp/teku/issues/18 (add seenTTL gossip parameter and its handling)

Teku integration PR: https://github.com/PegaSysEng/teku/pull/2983


src/main/kotlin/io/libp2p/etc/util/TimeLimitedMap.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val timestamps = LinkedList<Pair<Long, K>>()
+
+    override fun put(key: K, value: V): V? {
+        val oldVal = backMap[key]
+
+        if (oldVal == null) {
+            timestamps += timeSupplier() to key
+        }
+        backMap[key] = value
+        purge()
```
mbaxter, line 25: Don't the messages need to be purged on `get` as well?

mbaxter, line 25: Should we be purging first thing?  If the timeout is 1sec, and you `put(a,1)` then 1.5 seconds later you `put(a,1)`, shouldn't the value be set with a new timeout??

Nashatyrev: Replaced my custom class with Guava `Cache` implementation. This addresses both concerns

Nashatyrev: Regarding my implementation: I just made the minimal simplified class which suffices `seenMessage` requirements :)

mbaxter (approved):

src/main/kotlin/io/libp2p/etc/util/TimeLimitedMap.kt
mbaxter: makes sense - its just easier for long-term maintainability if classes function as expected independent of context

PR #150, opened 2020-10-28
Fix dependencies.

Nashatyrev:
- Include Noise-java as artifact instead of source dependency. Gradle source dependency is hardly managed as a transitive dependency.
- Netty and Protobuf classes are exposed with Libp2p API, so the dependency for these libs should be 'api' instead of 'implementation'


ajsutton (approved):
LGTM.

PR #155, opened 2020-11-09
Prepare for handling uncompressed Gossip MessageId

Nashatyrev:
Changes to Libp2p Gossip to make it possible for Teku to calculate Gossip `message-id` based on uncompressed payload content and allow to: 
- Avoid expensive uncompressing a gossip message twice (First on `message-id` calculation, second on message handling) 
- Avoid duplicate calculation of `message-id` (which need duplicate uncompressing) for duplicate inbound messages: by using `fast-message-id` (id based on compressed content)

The first is achieved by introducing a `PubsubMessage` wrapper class instead of the raw protobuf `Rpc.Message` and `AbstractRouter.messageFactory` by which the client code may create his own `PubsubMessage` implementation with necessary caching of uncompressed payload 

The second is achieved by changing `AbstractRouter.seenMessages` to a pluggable custom 'Map-like' `SeenCache` interface with `FastIdSeenCache` implementation 

Related Teku PR: https://github.com/ConsenSys/teku/pull/3188


ajsutton (approved):
LGTM. Nice work.

PR #161, opened 2020-11-16
Fix log ERROR when fail to decrypt inbound packet 

Nashatyrev:
We have failed to handle the case when `javax.crypto.AEADBadTagException` is the root cause of `CantDecryptInboundException`

Also `abruptlyClosing` flag was added to avoid nested decode exceptions during channel close. When channel is closing the netty `ByteToMessageDecoder` tries to feed all unprocessed bytes to the pipeline which again causes `CantDecryptInboundException` from security 'codec'. This makes sense when channel is closed remotely but should be prevented when we are closing the local channel abruptly

Fix when merged to Teku https://github.com/ConsenSys/teku/issues/3214

### DIFF
--- a/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
@@ -1,5 +1,7 @@
 package io.libp2p.etc.types
 
+import kotlin.reflect.KClass
+
 fun Boolean.whenTrue(run: () -> Unit): Boolean {
     if (this) {
         run()
@@ -36,3 +38,6 @@ fun <T> defer(f: (Deferrable) -> T): T {
         deferrable.execute()
     }
 }
+
+fun <T : Throwable> Throwable.hasCauseOfType(clazz: KClass<T>): Boolean =
+    clazz.isInstance(this) || (cause != null && cause!!.hasCauseOfType(clazz))
--- a/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
@@ -1,5 +1,6 @@
 package io.libp2p.etc.types
 
+import com.google.common.base.Throwables
 import kotlin.reflect.KClass
 
 fun Boolean.whenTrue(run: () -> Unit): Boolean {
@@ -39,5 +40,7 @@ fun <T> defer(f: (Deferrable) -> T): T {
     }
 }
 
-fun <T : Throwable> Throwable.hasCauseOfType(clazz: KClass<T>): Boolean =
-    clazz.isInstance(this) || (cause != null && cause!!.hasCauseOfType(clazz))
+fun <T : Throwable> Throwable.hasCauseOfType(clazz: KClass<T>) =
+    Throwables.getCausalChain(this)
+        .filter(clazz::isInstance)
+        .any()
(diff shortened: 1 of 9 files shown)


src/main/kotlin/io/libp2p/security/secio/SecIoCodec.kt
```diff
@@ -46,33 +49,42 @@ class SecIoCodec(val local: SecioParams, val remote: SecioParams) : MessageToMes
...
-            is SecureChannelError -> {
-                log.debug("Invalid SecIO content", cause)
-                ctx.channel().close()
-            }
-            else -> {
-                log.error("Unexpected error in SecIO channel", cause)
-            }
+        if (cause.hasCauseOfType(IOException::class)) {
+            // Trace level because having clients unexpectedly disconnect is extremely common
+            log.trace("IOException in Noise channel", cause)
```
mbaxter, line 74:
```suggestion
            log.trace("IOException in SecIO channel", cause)
```

src/main/kotlin/io/libp2p/security/secio/SecIoCodec.kt
```diff
@@ -46,33 +49,42 @@ class SecIoCodec(val local: SecioParams, val remote: SecioParams) : MessageToMes
...
+        } else if (cause.hasCauseOfType(SecureChannelError::class)) {
+            log.debug("Invalid Noise content", cause)
```
mbaxter, line 76:
```suggestion
            log.debug("Invalid SecIO content", cause)
```

src/main/kotlin/io/libp2p/security/secio/SecIoCodec.kt
```diff
@@ -46,33 +49,42 @@ class SecIoCodec(val local: SecioParams, val remote: SecioParams) : MessageToMes
...
+            closeAbruptly(ctx)
+        } else {
+            log.error("Unexpected error in Noise channel", cause)
```
mbaxter, line 79:
```suggestion
            log.error("Unexpected error in SecIO channel", cause)
```

ajsutton (approved):
LGTM.

src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
```diff
@@ -36,3 +38,6 @@ fun <T> defer(f: (Deferrable) -> T): T {
         deferrable.execute()
     }
 }
+
+fun <T : Throwable> Throwable.hasCauseOfType(clazz: KClass<T>): Boolean =
+    clazz.isInstance(this) || (cause != null && cause!!.hasCauseOfType(clazz))
```
mbaxter, line 43: Isn't it possible to get stuck in an infinite loop here?  In [teku we're using a util](https://github.com/ConsenSys/teku/blob/b7fae2614f5a4e9e0e0c50706e9ad6790e680357/util/src/main/java/tech/pegasys/teku/util/exceptions/ExceptionUtil.java#L22-L28) that handles recursive causes.

ajsutton:
And that just shows why I always like Meredith's reviews... :)

Nashatyrev:
> And that just shows why I always like Meredith's reviews... :)

Yep, Meredith is the top reviewer 🥇

src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
Nashatyrev: Yeah, good point. 
I heard something of such weird cases with looped `Throwable` causalities and both utils are handling those cases.

PR #164, opened 2020-11-19
Fix performance issues in the Gossip SeenCache

Nashatyrev:
Fix 3 performance issues which became visible under high Gossip load:

- `TTLCache.pruneOld()`  was very expensive: caaaa41
- `GossipRouter.handleIHave` sub-optimally calculated unseen messages with `seenMessages.messages.map { it.messageId }`: 
68de132 
- `FastIdSeenCache.remove` was very expensive: 8f5f993


src/main/kotlin/io/libp2p/etc/types/MutableBiMultiMap.kt
```diff
@@ -0,0 +1,86 @@
...
+operator fun <K, V> MutableBiMultiMap<K, V>.contains(key: K) = this[key] != null
+operator fun <K, V> MutableBiMultiMap<K, V>.minusAssign(key: K) = removeKey(key)
+
+internal class MutableBiMultiMapImpl<Key, Value> : MutableBiMultiMap<Key, Value> {
+    @VisibleForTesting
+    internal val keyToValue: MutableMap<Key, Value> = mutableMapOf()
+    @VisibleForTesting
+    internal val valueToKeys: MutableMap<Value, Set<Key>> = mutableMapOf()
+
+    override fun put(key: Key, value: Value) {
```
mbaxter, line 61: If there's an existing value for this key, don't we need to remove the existing mapping from oldValue -> key before we set the new value?

Nashatyrev: Eagle eye! Fixing this

mbaxter (approved):

PR #166, opened 2020-12-01
Fix unfair IWANT penalty

Nashatyrev:
Fix the following scenario: 
- Node receives unseen `IHAVE` for message `M` from peer `P1`
- Node sends `IWANT M` to `P1`
- Node receives message `M` from peer `P2` notify the client and waits for its validation 
- While still in process of validation the `M` is received  from `P1` (as a response to `IWANT`)
- `notifyAnyValidMessage` is not called for the latter inbound and pending entry is not removed from `iWantRequests`
- Peer `P1` is penalized for `IWANT` response timeout 

The `iWantRequests` entry need to be cleared on any message. If a message occurs invalid the peer would be penalized in another way 

Also add the test verifying that a long period of Gossip `IGNORE` (i.e. not publishing any messages) doesn't underscore the node with Eth2 params

### DIFF
--- a/src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
+++ b/src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
@@ -66,10 +66,10 @@ private val subnetTopicParams =
     (0..63).map { "$AttestTopicPrefix$it$AttestTopicSuffix" to Eth2DefaultAttestTopicParams }.toMap()
 val Eth2DefaultTopicsParams = GossipTopicsScoreParams(
     topicParams =
-    mapOf(
-        BlocksTopic to Eth2DefaultBlockTopicParams,
-        AggrAttestTopic to Eth2DefaultAggrAttestTopicParams
-    ) + subnetTopicParams
+        mapOf(
+            BlocksTopic to Eth2DefaultBlockTopicParams,
+            AggrAttestTopic to Eth2DefaultAggrAttestTopicParams
+        ) + subnetTopicParams
 )
 
 val Eth2DefaultPeerScoreParams = GossipPeerScoreParams(
--- a/src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
+++ b/src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
@@ -1,0 +1,96 @@
+package io.libp2p.pubsub.gossip
+
+import io.libp2p.etc.types.minutes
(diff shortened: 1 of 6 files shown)


src/test/kotlin/io/libp2p/pubsub/gossip/GossipPubsubRouterTest.kt
```diff
@@ -209,4 +213,50 @@ class GossipPubsubRouterTest : PubsubRouterTest({
             Assertions.assertFalse(testLogAppender.hasAnyWarns())
         }
     }
+
+    @Test
+    fun testIgnoreDoesntReduceScores() {
+        // check that with Eth2 Gossip scoring params
+        // a peers which IGNOREs all inbound messages doesn't got underscored
```
ajsutton, line 220: nit:
```suggestion
        // a peers which IGNOREs all inbound messages doesn't get underscored
```

src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
```diff
@@ -0,0 +1,96 @@
...
+// Spec constants
+val SlotDuration = 12.seconds
+val SlotsPerEpoch = 32
+val EpochDuration = SlotDuration * SlotsPerEpoch
+val BlocksTopic = "/eth2/00000000/beacon_block/ssz_snappy"
+val AggrAttestTopic = "/eth2/00000000/beacon_aggregate_and_proof/ssz_snappy"
+val AttestTopicPrefix = "/eth2/00000000/beacon_attestation_"
+val AttestTopicSuffix = "/ssz_snappy"
+
+val Eth2DefaultGossipParams = GossipParams(
```
ajsutton, line 16: I couldn't find the source for these params to confirm they are all set correctly.  But as I understand it at this point they're only used for the test - we'd have to specifically chose to use them on the Teku side before scoring is activated right?

We just need to be cautious about how we enable peer scoring - probably should have a CLI option to toggle it for a while so we can test it really carefully in real world conditions.

ajsutton (approved):
LGTM.

src/test/kotlin/io/libp2p/pubsub/gossip/Eth2GossipParams.kt
Nashatyrev: The source was some draft (I created this file a couple of months ago) from the group investigating Gossip params for Eth2. 
Right this is just for testing purposes for now and the precise numbers are not too important here. 
Yes, when enabling scoring in Teku we would need to revisit them for sure.

PR #168, opened 2020-12-03
Fix 2 flaky tests, clean up build 

Nashatyrev:
Fix 
- `SecureChannelTest` which periodically just hung the build: the reason was in using a regular `ArrayList` in multithreading scenario. Also discovered that channel methods were called from two threads: main and `TestChannel.executor`
- `HostTest.pingOverSecureConnection()` which periodically failed with non-expected exception: the reason was in unsynchronized `Ping` actions 
- Clean up build warnings


ajsutton (approved):
LGTM.

PR #173, opened 2021-01-25
Check causal chain for exceptions when deciding what log level to use

ajsutton:
Exceptions during decoding wind up being wrapped by netty in `DecoderException` so when deciding what level to use for exceptions, check the causal chain rather than just the outermost exception type.

See https://github.com/ConsenSys/teku/issues/3309 for an example of exceptions this would handle better.


Nashatyrev (approved):
LGTM

PR #175, opened 2021-02-03
Tweak gossip scoring config apis

mbaxter:
### Description
Update gossip scoring configuration API's:
* Expose setter for `GossipPeerScoreparamsBuilder.behaviourPenaltyThreshold`
* Add `Gossip.updateTopicScoreParams` to allow for topic scoring params to be updated after startup

### DIFF
--- a/src/main/kotlin/io/libp2p/etc/types/Delegates.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/Delegates.kt
@@ -31,7 +31,7 @@ fun cappedDouble(value: Double, decayToZero: Double = Double.MIN_VALUE): CappedV
  * and may drop value to [0.0] when the new value is less than [decayToZero]
  */
 fun cappedDouble(value: Double, decayToZero: Double = Double.MIN_VALUE, upperBound: () -> Double) =
-    CappedValueDelegate(value, { decayToZero }, { 0.0 }, upperBound, upperBound)
+    CappedValueDelegate(value, { decayToZero }, { 0.0 }, upperBound)
 
 // thanks to https://stackoverflow.com/a/47948047/9630725
 class LazyMutable<T>(val initializer: () -> T, val rejectSetAfterGet: Boolean = false) : ReadWriteProperty<Any?, T> {
--- a/src/main/kotlin/io/libp2p/etc/types/Delegates.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/Delegates.kt
@@ -17,14 +17,21 @@ fun <T> lazyVar(defaultValueInit: () -> T) = LazyMutable(defaultValueInit)
 fun <T> lazyVarInit(defaultValueInit: () -> T) = LazyMutable(defaultValueInit)
 
 fun <T : Comparable<T>> cappedVar(value: T, lowerBound: T, upperBound: T) =
-    CappedValueDelegate(value, lowerBound = lowerBound, upperBound = upperBound)
+    CappedValueDelegate(value, lowerBound = { lowerBound }, upperBound = { upperBound })
+
+/**
+ * Creates a Double delegate which may drop value to [0.0] when the new value is less than [decayToZero]
+ */
+fun cappedDouble(value: Double, decayToZero: Double = Double.MIN_VALUE): CappedValueDelegate<Double> {
+    return cappedDouble(value, decayToZero) { Double.MAX_VALUE }
+}
 
 /**
  * Creates a Double delegate which may cap upper bound (set [upperBound] when the new value is greater)
  * and may drop value to [0.0] when the new value is less than [decayToZero]
  */
-fun cappedDouble(value: Double, decayToZero: Double = Double.MIN_VALUE, upperBound: Double = Double.MAX_VALUE) =
-    CappedValueDelegate(value, decayToZero, 0.0, upperBound)
+fun cappedDouble(value: Double, decayToZero: Double = Double.MIN_VALUE, upperBound: () -> Double) =
+    CappedValueDelegate(value, { decayToZero }, { 0.0 }, upperBound, upperBound)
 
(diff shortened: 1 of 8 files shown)


src/main/kotlin/io/libp2p/etc/types/Delegates.kt
```diff
@@ -55,20 +60,20 @@ class LazyMutable<T>(val initializer: () -> T, val rejectSetAfterGet: Boolean =
...
+    private var value: C,
+    private val lowerBound: () -> C,
+    private val lowerBoundVal: () -> C = lowerBound,
+    private val upperBound: () -> C,
+    private val upperBoundVal: () -> C = upperBound
+) : ReadWriteProperty<Any?, C> {
 
     override fun getValue(thisRef: Any?, property: KProperty<*>): C {
+        val v1 = if (value > upperBound.invoke()) upperBoundVal.invoke() else value
+        value = if (value < lowerBound.invoke()) lowerBoundVal.invoke() else v1
```
Nashatyrev, line 72: Hm, modifying the `value` in getter looks slightly weird but it does the trick with dynamic bounds

src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
```diff
@@ -415,9 +415,22 @@ data class GossipPeerScoreParams(
  */
 class GossipTopicsScoreParams(
     private val defaultParams: GossipTopicScoreParams = GossipTopicScoreParams(),
-    private val topicParams: Map<Topic, GossipTopicScoreParams> = mapOf()
+    topicParamsMap: Map<Topic, GossipTopicScoreParams> = mapOf()
 ) {
+    val topicParams: MutableMap<Topic, GossipTopicScoreParams>
+    init {
+        this.topicParams = topicParamsMap.toMutableMap()
+    }
```
Nashatyrev, line 423: nit: 
```suggestion
    val topicParams: MutableMap<Topic, GossipTopicScoreParams>= topicParamsMap.toMutableMap()
```

src/main/kotlin/io/libp2p/etc/types/Delegates.kt
Nashatyrev, line 72: nit:
```suggestion
        val v1 = if (value > upperBound()) upperBoundVal() else value
        value = if (value < lowerBound()) lowerBoundVal() else v1
```

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -116,6 +117,14 @@ class GossipScore(
 
     private fun isInMesh(peer: P2PService.PeerHandler, topic: Topic) = getTopicScores(peer, topic).inMesh()
 
+    fun updateTopicParams(topicScoreParams: Map<String, GossipTopicScoreParams>) {
+        executor.execute {
```
Nashatyrev, line 121: nit: does it make sense just to make `topicsScoreParams.topicParams` map synchronized and call `setTopicParams` directly?

src/test/kotlin/io/libp2p/pubsub/gossip/GossipScoreTest.kt
```diff
@@ -33,26 +41,654 @@ class GossipScoreTest {
...
+            .setData(ByteString.copyFrom("Test", StandardCharsets.US_ASCII))
+            .build()
+    }
+
+    private fun mockPeer(): P2PService.PeerHandler {
+        val peerId = PeerId.random()
+        val peer = mockk<P2PService.PeerHandler>()
+        every { peer.peerId } returns peerId
+        every { peer.getIP() } returns "127.0.0.1"
+        return peer
```
Nashatyrev, line 692: Wow, titanic tests contribution! 👍

Nashatyrev (approved):
LGTM 👍 
Just some minor comments

src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
mbaxter: hmm - I can't seem to get that to compile

Nashatyrev: What's the error?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
mbaxter: Not sure - if we moved to a synchronized map, I think we can get cases where `calcTopicScore` performs part of the calculation with the old topic params and part of the calculation with new topic params.  Maybe that's okay?  But seems unexpected.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
mbaxter: Oh, weird - it was just a white space issue: 
Screen Shot 2021-02-04 at 11 23 22 AM

Working now 😄

src/main/kotlin/io/libp2p/etc/types/Delegates.kt
Nashatyrev: Yep, I saw this case and just could not invent anything better. 
> Which seems weird?

Just update the value in getter seems strange. 
Sorry if I maybe misused the word 'weird'. Not sure probably it sounds rather 'awful' than 'strange' :)

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: You are right it's better to update them atomically. Let's leave the `executor` then

PR #177, opened 2021-02-05
Release 0.7.0

mbaxter:
### Description

* Add the ability to pre-handle any inbound or outbound stream (#170)
* Check the full causal chain when handling exceptions (#173)
* Update gossip scoring APIs: allow topic scoring params to be updated after startup (#175)


Nashatyrev (approved):
LGTM

PR #179, opened 2021-02-24
Publish to CloudSmith

ajsutton:
Publish builds to cloudsmith instead of bintray.

Update instructions in README and add acknowledgement.


rolfyone (approved):
LGTM

PR #181, opened 2021-02-24
Fire readComplete() event to Stream channels

Nashatyrev:
It turned out that `readComplete` events are not propagated to child `Stream` channels and this caused `ReadTimeoutHandler` to behave incorrectly: it just didn't throw `TimeoutException` if at least one byte was read.

The idea is to propagate the `Connection` `readComplete` event to those `Streams` which were reading after the previous `readComplete` and are not yet closed.


ajsutton (approved):
LGTM.

PR #183, opened 2021-03-09
Release 0.8.0

ajsutton:
* Releases are published to cloudsmith instead of bintray
* readComplete events are correctly propagated to Stream channels


rolfyone (approved):
LGTM

PR #184, opened 2021-03-25
Optimize gossip score calculation

Nashatyrev:
## Description
On every inbound message `GossipRouter` checks the peer score against `graylistThreshold` to decide whether a message should be processed or ignored. 
The score calculation is not too cheap for extensive inbound message flow. 

The following optimizations were made: 
- Cache `PeerHandler.getIP()` 
- Cache topic score unless any of input values changed or time changed more than `timeInMeshQuantum`. Ideally that should be made on something like Rx `ObservableValue`s and `Bindings` to look nicer but that seems like an overkill for this change 
-  Maintain whitelist with expiration and message count for `acceptRequestsFrom` for peers with non-negative score to avoid score recalculation for every inbound message. The basic idea behind is that `graylistThreshold` should be far below `0` and if a peer has the score 0 or above we could unconditionally accept some reasonable amount of  its messages for say 1 second. 

## Performance 

Profiling graph of `AbstractRouter.onInbound()`

### Before
изображение

### After
изображение

### DIFF
--- a/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -55,7 +55,8 @@ open class GossipRouter @JvmOverloads constructor(
     subscriptionTopicSubscriptionFilter: TopicSubscriptionFilter = TopicSubscriptionFilter.AllowAllTopicSubscriptionFilter()
 ) : AbstractRouter(subscriptionTopicSubscriptionFilter, params.maxGossipMessageSize) {
 
-    val acceptRequestsWhitelistThreshold = 0
+    val acceptRequestsWhitelistThresholdScore = 0
+    val acceptRequestsWhitelistMaxMessages = 128
     val acceptRequestsWhitelistDuration = 1.seconds
     val score by lazy { GossipScore(scoreParams, executor, curTimeMillis) }
     val fanout: MutableMap<Topic, MutableSet<PeerHandler>> = linkedMapOf()
@@ -76,7 +77,7 @@ open class GossipRouter @JvmOverloads constructor(
             TimeUnit.MILLISECONDS
         )
     }
-    private val acceptRequestsWhitelist = mutableMapOf<PeerHandler, Long>()
+    private val acceptRequestsWhitelist = mutableMapOf<PeerHandler, AcceptRequestsWhitelistEntry>()
 
     override val seenMessages: SeenCache<Optional<ValidationResult>> by lazy {
         TTLSeenCache(SimpleSeenCache(), params.seenTTL, curTimeMillis)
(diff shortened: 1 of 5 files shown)


src/test/kotlin/io/libp2p/etc/types/DelegatesTest.kt
```diff
@@ -14,8 +14,7 @@ class DelegatesTest {
 
     var cappedValueDelegate: Double by CappedValueDelegate(0.0, { min.get() }, { minVal.get() }, { max.get() }, { maxVal.get() })
     var cappedInt: Int by cappedVar(10, 5, 20)
-    var cappedDouble: Double by cappedDouble(0.0, 1.0) { max.get() }
-
+    var cappedDouble: Double by cappedDouble(0.0, 1.0, { -> max.get() })
     @Test
```
mbaxter, line 18: Might be worth a couple tests to check the update callback is invoked correctly

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -59,6 +55,9 @@ open class GossipRouter @JvmOverloads constructor(
     subscriptionTopicSubscriptionFilter: TopicSubscriptionFilter = TopicSubscriptionFilter.AllowAllTopicSubscriptionFilter()
 ) : AbstractRouter(subscriptionTopicSubscriptionFilter, params.maxGossipMessageSize) {
 
+    val acceptRequestsWhitelistThresholdScore = 0
+    val acceptRequestsWhitelistMaxMessages = 128
+    val acceptRequestsWhitelistDuration = 1.seconds
```
mbaxter, line 60: Any particular reason to choose these specific params?  Might be worth a comment.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -173,7 +174,28 @@ open class GossipRouter @JvmOverloads constructor(
...
+
+            acceptRequestsWhitelist[peer] = whitelistEntry.incrementMessageCount()
+            return true
+        }
+
+        val peerScore = score.score(peer)
+        if (peerScore >= acceptRequestsWhitelistThresholdScore) {
+            acceptRequestsWhitelist[peer] =
+                AcceptRequestsWhitelistEntry(curTime + acceptRequestsWhitelistDuration.toMillis())
+        }
```
mbaxter, line 196: Should we clear out the whitelist entry otherwise?

mbaxter (approved):

src/test/kotlin/io/libp2p/etc/types/DelegatesTest.kt
Nashatyrev: Done: aaca0d0c

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: The idea was that those param values
- from one side are pretty small and safe: peer unlikely be able to drop its score to `graylist` with 128 messages. But even if so then it's not critical to accept some extra messages before blocking - not too much space for DoS here
- from the other side param values are pretty high to yield good performance gain

Will add these comments to the code

src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
Nashatyrev: Yep, makes sense 👍 
c545f19

PR #188, opened 2021-04-10
Add method to retrieve peer score by peerId

BogdanStirbat:
Added a method to retrieve peer score by peerId. This is needed for [making gossip scoring information visible to users](https://github.com/ConsenSys/teku/issues/3557).

### DIFF
--- a/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
@@ -3,6 +3,7 @@ package io.libp2p.pubsub.gossip
 import io.libp2p.core.Connection
 import io.libp2p.core.ConnectionHandler
 import io.libp2p.core.P2PChannel
+import io.libp2p.core.PeerId
 import io.libp2p.core.Stream
 import io.libp2p.core.multistream.ProtocolBinding
 import io.libp2p.core.multistream.ProtocolDescriptor
@@ -21,6 +22,10 @@ class Gossip @JvmOverloads constructor(
 
     fun updateTopicScoreParams(scoreParams: Map<String, GossipTopicScoreParams>) {
         router.score.updateTopicParams(scoreParams)
+    }
+
+    fun getGossipScore(peerId: PeerId): Double {
+        return router.score.getCachedScore(peerId)
     }
 
     override val protocolDescriptor =
--- a/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
@@ -3,7 +3,6 @@ package io.libp2p.pubsub.gossip
 import io.libp2p.core.Connection
 import io.libp2p.core.ConnectionHandler
 import io.libp2p.core.P2PChannel
-import io.libp2p.core.PeerId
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -193,6 +193,10 @@ class GossipScore(
         }
     }
 
+    fun getPeerScore(peerId: PeerId): PeerScores {
+        return peerScores.computeIfAbsent(peerId) { PeerScores() }
```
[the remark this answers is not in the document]
BogdanStirbat: I was a bit confused, indeed. I'm sorry about that. Now, the [score](https://github.com/libp2p/jvm-libp2p/blob/6cac516d2ea5f75d96dcc4f0167ff0affd94fd6f/src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt#L163) function puts the result in a cache; `getCachedScore(peerId: PeerId): Double` will return what's in the cache.

src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
```diff
@@ -24,10 +23,6 @@ class Gossip @JvmOverloads constructor(
         router.score.updateTopicParams(scoreParams)
     }
 
-    fun getGossipScore(peerId: PeerId): GossipScore.PeerScores {
```
mbaxter: I think we still want to expose this from `Gossip`

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -132,6 +132,7 @@ class GossipScore(
     val topicParams = params.topicsScoreParams
 
     private val validationTime: MutableMap<PubsubMessage, Long> = createLRUMap(1024)
+    val cachedScores = mutableMapOf<PeerId, Double>()
```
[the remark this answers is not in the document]
BogdanStirbat: I understand. I made the modifications.

src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
BogdanStirbat: I see, the API will be exported from Gossip.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -117,6 +118,7 @@ class GossipScore(
         val ips = mutableSetOf<String>()
         var connectedTimeMillis: Long = 0
         var disconnectedTimeMillis: Long = 0
+        var cachedScore: AtomicReference<Double> = AtomicReference(0.0)
```
mbaxter, line 121:
```suggestion
        val cachedScore: AtomicReference<Double> = AtomicReference(0.0)
```

mbaxter (commented):
Looks good!  Left one small nit.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
BogdanStirbat: Yes, you're right, cachedScore is a val.

PR #189, opened 2021-04-22
Release 0.8.2

mbaxter:
### Description
* Expose `Gossip.getGossipScore(peerId)` method for retrieving a peer's gossip score (#188)

### DIFF
--- a/src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
@@ -13,6 +13,7 @@ import java.util.Optional
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
 import kotlin.math.max
 import kotlin.math.min
 import kotlin.math.pow
@@ -117,6 +118,7 @@ class GossipScore(
         val ips = mutableSetOf<String>()
         var connectedTimeMillis: Long = 0
         var disconnectedTimeMillis: Long = 0
+        val cachedScore: AtomicReference<Double> = AtomicReference(0.0)
 
         val topicScores = mutableMapOf<Topic, TopicScores>()
         var behaviorPenalty: Double by cappedDouble(0.0, peerParams.decayToZero)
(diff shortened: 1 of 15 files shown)


src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -117,6 +118,7 @@ class GossipScore(
         val ips = mutableSetOf<String>()
         var connectedTimeMillis: Long = 0
         var disconnectedTimeMillis: Long = 0
+        val cachedScore: AtomicReference<Double> = AtomicReference(0.0)
```
Nashatyrev, line 121: - `AtomicDouble` could be used here 
- All the gossip structures are not thread safe by design (they are accessed from a single thread). Accessing `peerScores[peerId]` would most likely work thought is still not thread safe. With respect to this I'm not sure that using `Atomic*` makes too much sense here. Probably a simple `var cachedScore: Double` would be enough?

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
```diff
@@ -193,6 +197,10 @@ class GossipScore(
         }
     }
 
+    fun getCachedScore(peerId: PeerId): Double {
```
Nashatyrev, line 200: I'd better return `Optional<Double>` here since `0.0` score is not the same as absence of score

mbaxter: Isn't 0 a reasonable default?  This should only show up in the brief window between the time a peer connects and we requests its score.

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
mbaxter: Makes sense - looks like we should just make `peerScores` a `ConcurrentHashMap`, and use a plain `Double` for the cached value.  WDYT?

mbaxter:
Closing this for now so we can address your points @Nashatyrev

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: Agree. I was actually thinking about the case when score requested for unknown peer. But `0.0` default indeed looks fine as if/when a peer is connected it gets `0.0` score by default. Se we may assume any unknown peer has score `0.0`

src/main/kotlin/io/libp2p/pubsub/gossip/GossipScore.kt
Nashatyrev: Yeah makes sense. I'd also mark it with `@Volatile` to be safe 
Alternatively we may consider a dedicated `ConcurrentHashMap` for the score cache

PR #190, opened 2021-04-23
Safely access cached gossip score across threads

mbaxter:
### Description

Rework access to cached gossip score value.


Nashatyrev (approved):
LGTM

PR #191, opened 2021-04-23
Release 0.8.2

mbaxter:
### Description
* Expose `Gossip.getGossipScore(peerId)` method for retrieving a peer's gossip score (#188)
* Rework access to cached gossip score value (#190)


Nashatyrev (approved):
LGTM

PR #193, opened 2021-06-24
Add a timeout to the connection upgrade process

ajsutton:
Addresses an issue where connections may leak if the secure channel or muxer establishment process stalls (e.g. because the remote peer doesn't respond). This is done using a rather basic timeout which appears to solve the issue but may not be the most elegant solution.  If the upgrade fails, the channel is closed to prevent the connection leaking.

At minimum we need to make the timeout configurable but it may be that this timeout should be pushed down into the individual handlers (eg `NoiseXXSecureChannel` and the actual muxing implementations).

### DIFF
--- a/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -271,7 +271,7 @@ abstract class AbstractRouter(
                     when {
                         err != null -> logger.warn("Exception while handling message from peer $peer: ${it.first}", err)
                         res == ValidationResult.Invalid -> logger.debug("Invalid pubsub message from peer $peer: ${it.first}")
-                        res == ValidationResult.Ignore -> logger.debug("Ingnoring pubsub message from peer $peer: ${it.first}")
+                        res == ValidationResult.Ignore -> logger.trace("Ignoring pubsub message from peer $peer: ${it.first}")
                         else -> {
                             newValidatedMessages(singletonList(it.first), peer)
                             flushAllPending()
--- a/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
+++ b/src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
@@ -25,6 +25,7 @@ import io.netty.channel.CombinedChannelDuplexHandler
 import io.netty.channel.SimpleChannelInboundHandler
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder
 import io.netty.handler.codec.LengthFieldPrepender
+import io.netty.handler.timeout.ReadTimeoutHandler
 import org.apache.logging.log4j.LogManager
 import spipe.pb.Spipe
 import java.util.concurrent.CompletableFuture
@@ -33,8 +34,10 @@ private enum class Role(val intVal: Int) { INIT(HandshakeState.INITIATOR), RESP(
 
 private val log = LogManager.getLogger(NoiseXXSecureChannel::class.java)
 private const val HandshakeNettyHandlerName = "HandshakeNettyHandler"
+private const val HandshakeReadTimeoutNettyHandlerName = "HandshakeReadTimeoutNettyHandler"
 private const val NoiseCodeNettyHandlerName = "NoiseXXCodec"
 private const val MaxCipheredPacketLength = 65535
+private const val HandshakeTimeoutSec = 5
 
 class UShortLengthCodec : CombinedChannelDuplexHandler<LengthFieldBasedFrameDecoder, LengthFieldPrepender>(
     LengthFieldBasedFrameDecoder(MaxCipheredPacketLength + 2, 0, 2, 0, 2),
@@ -70,7 +73,8 @@ class NoiseXXSecureChannel(private val localKey: PrivKey) :
         val handshakeComplete = CompletableFuture<SecureChannel.Session>()
         // Packet length codec should stay forever.
         ch.pushHandler(UShortLengthCodec())
-        // Handshake handle is to be removed when handshake is complete
+        // Handshake and ReadTimeout handlers are to be removed when handshake is complete
+        ch.pushHandler(HandshakeReadTimeoutNettyHandlerName, ReadTimeoutHandler(HandshakeTimeoutSec))
         ch.pushHandler(
             HandshakeNettyHandlerName,
             NoiseIoHandshake(localKey, handshakeComplete, if (ch.isInitiator) Role.INIT else Role.RESP)
@@ -302,6 +306,7 @@ private class NoiseIoHandshake(
             "NoisePacketSplitter",
             SplitEncoder(MaxCipheredPacketLength - session.aliceCipher.macLength)
         )
+        ctx.pipeline().remove(HandshakeReadTimeoutNettyHandlerName)
         ctx.pipeline().remove(this)
         ctx.fireChannelActive()
     } // handshakeSucceeded
@@ -313,6 +318,7 @@ private class NoiseIoHandshake(
         log.debug("Noise handshake failed", cause)
 
         handshakeComplete.completeExceptionally(cause)
+        ctx.pipeline().remove(HandshakeReadTimeoutNettyHandlerName)
         ctx.pipeline().remove(this)
     }
 } // class NoiseIoHandshake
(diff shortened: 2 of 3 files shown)


Nashatyrev (approved):
The usual Netty way of handling read timeouts is adding `ReadTimeoutHandler` to the pipeline. For some reason I thought all individual handlers add it when appropriate, but was surprised that only `multistream.Negotiator` does. 

Your fix however also should work well 👍 

>  may be that this timeout should be pushed down into the individual handlers (eg `NoiseXXSecureChannel` and the actual muxing implementations).

Yep timeout handler should definitely be added to `NoiseXXSecureChannel` for negotiation period.  Though I think that the connection should be kept alive after Noise negotiation (and also at mux handler) despite nothing is received. At this stage I believe the read timeout should be handled at application level.

ajsutton:
The connection handler isn't notified of the connection until after the noise negotiation completes so the application doesn't know anything about the connection yet. 

So we'd probably have to add a method to the connection handler that is used to report these session establishment failures. Will have to dig into the code properly tomorrow though to get a better understanding of how it fits together.

Nashatyrev:
Yep, handling read timeouts on app level are performed after successful handshake only (this was just a general comment and  is not directly related to the issue observed)

There are handshake steps which should be aborted due to read timeout 
1. `Multiprotocol`: where parties agreed on security protocol (e.g. `/noise`) This timeout handler should take care of it: https://github.com/libp2p/jvm-libp2p/blob/ec43a5be56b5157c9e6223cf3e2da699b10d51a1/src/main/kotlin/io/libp2p/multistream/Negotiator.kt#L75
2. 'NoiseIoHandshake`: Noise handshake. This handler is added synchronously when the previous step succeeds. Timeout handler is missing here. Potential fix is below
3. `Multiprotocol` (again): where parties agreed on muxer. The same Timeout handler as in (1)
4. At this stage the Muxer is set up and now its up to application how to handle read  timeouts. 

```patch
Index: src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt
IDEA additional info:
<+>UTF-8
--- src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt	(revision c545f198abf04689a2614b40e196eed282d1e1e8)
+++ src/main/kotlin/io/libp2p/security/noise/NoiseXXSecureChannel.kt	(date 1605081438187)
@@ -25,6 +25,7 @@
 import io.netty.channel.SimpleChannelInboundHandler
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder
 import io.netty.handler.codec.LengthFieldPrepender
+import io.netty.handler.timeout.ReadTimeoutHandler
 import org.apache.logging.log4j.LogManager
 import spipe.pb.Spipe
@@ -33,8 +34,10 @@
 
 private val log = LogManager.getLogger(NoiseXXSecureChannel::class.java)
 private const val HandshakeNettyHandlerName = "HandshakeNettyHandler"
+private const val HandshakeReadTimeoutNettyHandlerName = "HandshakeReadTimeoutNettyHandler"
 private const val NoiseCodeNettyHandlerName = "NoiseXXCodec"
 private const val MaxCipheredPacketLength = 65535
+private const val HandshakeTimeoutSec = 5
 
 class UShortLengthCodec : CombinedChannelDuplexHandler<LengthFieldBasedFrameDecoder, LengthFieldPrepender>(
     LengthFieldBasedFrameDecoder(MaxCipheredPacketLength + 2, 0, 2, 0, 2),
@@ -70,7 +73,8 @@
         val handshakeComplete = CompletableFuture<SecureChannel.Session>()
         // Packet length codec should stay forever.
         ch.pushHandler(UShortLengthCodec())
-        // Handshake handle is to be removed when handshake is complete
+        // Handshake and ReadTimeout handlers are to be removed when handshake is complete
+        ch.pushHandler(HandshakeReadTimeoutNettyHandlerName, ReadTimeoutHandler(HandshakeTimeoutSec))
         ch.pushHandler(
             HandshakeNettyHandlerName,
             NoiseIoHandshake(localKey, handshakeComplete, if (ch.isInitiator) Role.INIT else Role.RESP)
@@ -302,6 +306,7 @@
             "NoisePacketSplitter",
             SplitEncoder(MaxCipheredPacketLength - session.aliceCipher.macLength)
         )
+        ctx.pipeline().remove(HandshakeReadTimeoutNettyHandlerName)
         ctx.pipeline().remove(this)
         ctx.fireChannelActive()
     } // handshakeSucceeded
@@ -313,6 +318,7 @@
         log.debug("Noise handshake failed", cause)
 
         handshakeComplete.completeExceptionally(cause)
+        ctx.pipeline().remove(HandshakeReadTimeoutNettyHandlerName)
         ctx.pipeline().remove(this)
     }
 } // class NoiseIoHandshake
```

Nashatyrev:
Hmm just realized from your logs in slack that Noise actually completes its handshake. Then is looks that something else is stuck 
```
{"timestamp":"2021-06-24T02:00:38,104","level":"DEBUG","thread":"nioEventLoopGroup-3-2","class":"Firewall","message":"Firewall added for incoming connection from /58.33.72.226:16125","throwable":""}
{"timestamp":"2021-06-24T02:00:38,104","level":"DEBUG","thread":"nioEventLoopGroup-3-2","class":"NoiseXXSecureChannel","message":"Starting handshake","throwable":""}
{"timestamp":"2021-06-24T02:00:38,116","level":"DEBUG","thread":"nioEventLoopGroup-3-1","class":"NoiseXXSecureChannel","message":"Noise handshake READ_MESSAGE","throwable":""}
{"timestamp":"2021-06-24T02:00:38,117","level":"DEBUG","thread":"nioEventLoopGroup-3-1","class":"NoiseXXSecureChannel","message":"Verifying noise static key payload","throwable":""}
{"timestamp":"2021-06-24T02:00:38,117","level":"DEBUG","thread":"nioEventLoopGroup-3-1","class":"NoiseXXSecureChannel","message":"Remote verification is true","throwable":""}
{"timestamp":"2021-06-24T02:00:38,117","level":"DEBUG","thread":"nioEventLoopGroup-3-1","class":"NoiseXXSecureChannel","message":"Split complete","throwable":""}
```

Nashatyrev:
If the case is reproducible, could be helpful to turn on wire logs and see what exactly happens there: 
```
--Xlog-wire-cipher-enabled 
--Xlog-wire-plain-enabled
```

ajsutton:
It's reproducible in the sense that if you run a node for an hour or so it will happen but I don't yet know how to reproduce it manually.

The other thing that happens in the timeout I added is negotiation of the muxer so possibly not getting a response to that step but you suggest there's a timeout on that already. And there's a possibility that an exception is being thrown somewhere.  From what I can see, any exception in this process will be forwarded to the `connectionEstablished` future (https://github.com/libp2p/jvm-libp2p/blob/4da53240a905d0caeb780011074f2569b2d8b960/src/main/kotlin/io/libp2p/transport/implementation/ConnectionBuilder.kt#L26) but that's only used for outgoing connections - as far as I can tell, incoming connections would silently ignore those errors.

I've started a build running without the timeout added here, but log messages to indicate when the secure session and muxer are setup correctly and logging any exceptions that occur anywhere in the process.  Hopefully that will narrow down the cause a bit further.

ajsutton:
So got one case:
```
{"timestamp":"2021-06-24T23:18:05,991","level":"DEBUG","thread":"nioEventLoopGroup-3-4","class":"ConnectionBuilder","message":"Building connection with /119.254.81.66:59736","throwable":""}
{"timestamp":"2021-06-24T23:18:05,991","level":"DEBUG","thread":"nioEventLoopGroup-3-4","class":"NoiseXXSecureChannel","message":"Starting handshake","throwable":""}
```

and that's the last we hear of it. The Noise setup never completes. I assume the read timeout handler will catch that case though, so will add that in now and see if we still have problems.

ajsutton:
Been running for a couple of hours with the read timeout handler added to noise and not seen any leaked connections yet - even without closing the connection in an exception handler.  I'll give it a while longer but it looks like that may be the only fix that's required.  The logs can be misleading because there's lots of connections going on and the noise logs don't provide any indication of which client they're negotiating with at the time.

ajsutton:
I've run with just the ReadTimeoutListener added for noise most of the day now and there are no leaked connections.  So I think it's worth merging that and as far as I can tell it fixes the issue. It's definitely an improvement and a good thing to do.  We'll get the fix through into Teku and monitor over a longer period to see if further fixes are required.

PR #196, opened 2021-09-08
Throw specialized exception when no suitable peers to broadcast outbound exception

Nashatyrev:
Throw specialized exception when no suitable peers to broadcast outbound exception. 
Currently an abstract `io.libp2p.etc.types.NothingToCompleteException` is thrown. 
The final intention is to get rid of noisy DEBUG logs with stacktrace in Teku client


ajsutton (approved):
LGTM.

PR #204, opened 2021-12-20
Improve multistream implementation 

Nashatyrev:
- Limit the length of a single protocol id string to 1024
- Limit the total time spend for a protocol negotiation


ajsutton (approved):
LGTM.

PR #205, opened 2021-12-21
A couple of Mplex fixes

Nashatyrev:
- Check mplex frame length prior to reading all frame content. If the length exceeds max bound then discard remaining data and close connection 
- Slice outbound data if it exceeds max mplex frame data length (worth to be fixed in the light of increased message size limit up to 10Mb in the Merge fork)


ajsutton (approved):
LGTM.

PR #207, opened 2022-01-12
Update build tools.

ajsutton:
Move jcenter repo after mavenCentral.  Will need https://github.com/ConsenSys/noise-java/pull/4 merged and pull in a new version before we can actually drop JCenter.


ajsutton:
Sigh. We don't cache dependencies so can't build until we get that noise-java update or jcenter comes back.

rolfyone (approved):
LGTM

PR #215, opened 2022-01-24
Fix remote PeerId check on dial 

Nashatyrev:
- Some `Multiaddr` class refactorings 
- Verify remote `PeerId` on connect


ajsutton (approved):
LGTM.

PR #217, opened 2022-01-25
Multiaddr refactor

Nashatyrev:
- Refactor `Multiaddr` to be represented by a list of `MultiaddrComponent` instead of `Pair`
- Refactor `Protocol` enum implementation. Add component value validation
- Refactor/clean up `Multiaddr` methods/constructors 
- Piggy tailing minor `NoiseXXSecureChannel` update to make it more extendable 

Note: breaking API changes


src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -166,3 +103,100 @@ enum class Protocol(val code: Int, val size: Int, val typeName: String) {
...
+private val PEER_ID_VALIDATOR: (Protocol, ByteArray?) -> Unit = { protocol, bytes ->
+    if (bytes == null)
+        throw IllegalArgumentException("Non-null value expected for PeerId in $protocol")
+    PeerId(bytes) // constructor validates array size
+}
+
+private val NO_PARSER: (Protocol, String) -> ByteArray = { protocol, _ ->
+    throw IllegalArgumentException("No value serializer for protocol $protocol")
+}
+private val NO_DESERIALIZER: (Protocol, ByteArray) -> String = { protocol, _ ->
```
ajsutton, line 132: I'm confused about the PARSER vs DESERIALIZER here.  In my mind a PARSER goes from String to internal representation and a SERIALIZER goes from internal to String.  A PARSER and DESERIALIZER would then be the same thing.

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -166,3 +103,100 @@ enum class Protocol(val code: Int, val size: Int, val typeName: String) {
...
+    byteBuf(2).writeShort(x).toByteArray()
+}
+private val UINT16_DESERIALIZER: (Protocol, ByteArray) -> String = { _, bytes ->
+    bytes.toByteBuf().readUnsignedShort().toString()
+}
+private val UTF8_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
+    addr.toByteArray(StandardCharsets.UTF_8)
+}
+private val UTF8_DESERIALIZER: (Protocol, ByteArray) -> String = { _, bytes ->
+    String(bytes, StandardCharsets.UTF_8)
```
ajsutton, line 166: I don't think this will reject invalid characters (just replace them with `?`).  If we want to fully validate we could use `StandardCharsets.UTF_8.newDecoder().onMalformedInput(CodingErrorAction.REPORT).onUnmappableCharacter(CodingErrorAction.REPORT).decode()` but that adds a bunch of complexity and I'm not entirely sure it's worth it.

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -166,3 +103,100 @@ enum class Protocol(val code: Int, val size: Int, val typeName: String) {
...
+        throw IllegalArgumentException("The address is not IPv4 address: $addr")
+    }
+    inetAddr.address
+}
+private val IP4_DESERIALIZER: (Protocol, ByteArray) -> String = { _, bytes ->
+    Inet4Address.getByAddress(bytes).toString().drop(1)
+}
+private val IP6_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
+    Inet6Address.getByName(addr).address
+}
```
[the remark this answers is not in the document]
Nashatyrev: Good point 👍

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -166,3 +103,100 @@ enum class Protocol(val code: Int, val size: Int, val typeName: String) {
...
+    throw IllegalArgumentException("No value serializer for protocol $protocol")
+}
+private val NO_DESERIALIZER: (Protocol, ByteArray) -> String = { protocol, _ ->
+    throw IllegalArgumentException("No value deserializer for protocol $protocol")
+}
+private val IP4_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
+    if (!InetAddresses.isInetAddress(addr)) {
+        throw IllegalArgumentException("Couldn't parse IPv4 IP: `$addr`")
+    }
+    val inetAddr = Inet4Address.getByName(addr)
```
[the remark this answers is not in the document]
Nashatyrev: Good point 👍

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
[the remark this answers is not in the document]
Nashatyrev: Fixed: 655a9d2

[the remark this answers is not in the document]
Nashatyrev: Also changed the IPv6 'deserializer': 4c5bc5a

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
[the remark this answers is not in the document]
Nashatyrev: Fixed: 655a9d2

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
Nashatyrev: Right, better be safe. However I've added a UTF-8 validator instead: 6375f2a

src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
Nashatyrev: Oh, that was also breaking my mind. 
I used to term 'serializer/deserializer' for converting 'bytes' from/to internal representation
'parser' in my mind is 'string => internal representation' converter. 
If we assume that 'bytes' is internal form, then it should be something opposite to 'parser'. 'Stringifier', 'Composer', 'Unparser'? What term do you think is the most adequate? 

Just don't want to confuse it with the 'binary serializer'. If take the `Multiaddr` class for example then it clearly has 3 representations: binary, string and internal. And they are all distinct

Nashatyrev: In other words: what should be the name of an substance who does `toString()`? 😄

ajsutton: oof 3 meaningful states makes it hard. While it's not valid English, `Stringifier` is probably a good name as it's very clear what it does.  Or possibly parser becomes `STRING_TO_BYTES` and the current deserialiser becomes `BYTES_TO_STRING`?

ajsutton (approved):
LGTM. Just the naming of parsers/deserializers/whatevers - not sure there's a "right" answer there so go with what you think and we can always change it if we think of something better later.

PR #221, opened 2022-02-17
Change hardcoded sleep to waiter pattern

jtraglia:
There was a test that was failing on my laptop that was due to a timed sleep that was not long enough. Changing the wait to 20 seconds would fix the problem, but that makes the build slow for everyone. This will bail within a second of acknowledging all of the connections. If it hasn't acknowledged all of the connections within 30 seconds, the test will fail.


ajsutton (approved):
LGTM.

PR #226, opened 2022-02-28
Forbid negative uvarint values

### LINKED ISSUE #225: Unsigned varints allow negative values
### The Problem

I think I found a minor issue in `readUvarint` functions, which convert [uvarint](https://github.com/multiformats/unsigned-varint) buffers to `Long` values. I'm able to provide values that result in a large negative number, via integer overflow. I found the implementation interesting because the spec states that uvarints have a max of 9 bytes, but the loop is from `for (i in 0..9)` (10 bytes).

> For security, to avoid memory attacks, we use a "practical max" of 9 bytes. Though there is no theoretical limit, and future specs can grow this number if it is truly necessary to have code or length values equal to or larger than 2^63

https://github.com/libp2p/jvm-libp2p/blob/811043429b1e8540a1fa3c1bb38e6b01a0a28d79/src/main/kotlin/io/libp2p/etc/types/ByteBufExt.kt#L27-L49

https://github.com/libp2p/jvm-libp2p/blob/511c5a2f03c8401f658be15667326fd5e9dc8168/src/main/kotlin/io/libp2p/etc/types/ByteArrayExt.kt#L73-L98

I wrote a simple [proof of concept in Kotlin Playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjEwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiQGtvdGxpbi5FeHBlcmltZW50YWxVbnNpZ25lZFR5cGVzXG5mdW4gQnl0ZUFycmF5LnJlYWRVdmFyaW50KCk6IFBhaXI8TG9uZywgQnl0ZUFycmF5Pj8ge1xuICAgIHZhciB4OiBMb25nID0gMFxuICAgIHZhciBzID0gMFxuXG4gICAgdmFyIGluZGV4ID0gMFxuICAgIHZhciByZXN1bHQ6IExvbmc/ID0gbnVsbFxuICAgIGZvciAoaSBpbiAwLi45KSB7XG4gICAgICAgIHZhbCBiID0gdGhpcy5nZXQoaW5kZXgrKykudG9VQnl0ZSgpLnRvU2hvcnQoKVxuICAgICAgICBpZiAoYiA8IDB4ODApIHtcbiAgICAgICAgICAgIGlmIChpID09IDkgJiYgYiA+IDEpIHtcbiAgICAgICAgICAgICAgICByZXR1cm4gbnVsbFxuICAgICAgICAgICAgfVxuICAgICAgICAgICAgcmVzdWx0ID0geCBvciAoYi50b0xvbmcoKSBzaGwgcylcbiAgICAgICAgICAgIGJyZWFrXG4gICAgICAgIH1cbiAgICAgICAgeCA9IHggb3IgKGIudG9Mb25nKCkgYW5kIDB4N2Ygc2hsIHMpXG4gICAgICAgIHMgKz0gN1xuICAgIH1cblx0XG4gICAgLy8gSSB0aGluayB5b3UgbWVhbnQgdG8gY2hlY2sgaW5kZXggaGVyZTpcbiAgICAvLyBpZiAocmVzdWx0ICE9IG51bGwgJiYgcmVzdWx0IDw9IHNpemUpIHtcbiAgICBpZiAocmVzdWx0ICE9IG51bGwgJiYgaW5kZXggPD0gc2l6ZSkge1xuICAgICAgICByZXR1cm4gUGFpcihyZXN1bHQsIHNsaWNlKEludFJhbmdlKGluZGV4LCBzaXplIC0gMSkpLnRvQnl0ZUFycmF5KCkpXG4gICAgfVxuXG4gICAgcmV0dXJuIG51bGxcbn1cblxuLy8gSGVscGVyIGZ1bmN0aW9uIGZvciBjb252ZXJ0aW5nIGludHMgdG8gYnl0ZSBhcnJheVxuZnVuIGJ5dGVBcnJheU9mSW50cyh2YXJhcmcgaW50czogSW50KSA9IEJ5dGVBcnJheShpbnRzLnNpemUpIHsgcG9zIC0+IGludHNbcG9zXS50b0J5dGUoKSB9XG5cbmZ1biBtYWluKCkge1xuICAgXHQvLyAyXjYzIGlzIDksMjIzLDM3MiwwMzYsODU0LDc3NSw4MDhcbiAgICB2YWwgYXJyOiBCeXRlQXJyYXkgPSBieXRlQXJyYXlPZkludHMoXG4gICAgICAgIEludGVnZXIucGFyc2VJbnQoXCIxMDAwMDAwMFwiLCAyKSwgLy8gMVxuICAgICAgICBJbnRlZ2VyLnBhcnNlSW50KFwiMTAwMDAwMDBcIiwgMiksIC8vIDJcbiAgICAgICAgSW50ZWdlci5wYXJzZUludChcIjEwMDAwMDAwXCIsIDIpLCAvLyAzXG4gICAgICAgIEludGVnZXIucGFyc2VJbnQoXCIxMDAwMDAwMFwiLCAyKSwgLy8gNFxuICAgICAgICBJbnRlZ2VyLnBhcnNlSW50KFwiMTAwMDAwMDBcIiwgMiksIC8vIDVcbiAgICAgICAgSW50ZWdlci5wYXJzZUludChcIjEwMDAwMDAwXCIsIDIpLCAvLyA2XG4gICAgICAgIEludGVnZXIucGFyc2VJbnQoXCIxMDAwMDAwMFwiLCAyKSwgLy8gN1xuICAgICAgICBJbnRlZ2VyLnBhcnNlSW50KFwiMTAwMDAwMDBcIiwgMiksIC8vIDhcbiAgICAgICAgSW50ZWdlci5wYXJzZUludChcIjEwMDAwMDAwXCIsIDIpLCAvLyA5XG4gICAgICAgIEludGVnZXIucGFyc2VJbnQoXCIwMDAwMDAwMVwiLCAyKSAgLy8gMTBcbiAgICApXG4gICAgcHJpbnRsbihhcnIucmVhZFV2YXJpbnQoKSlcbn0ifQ==) to demonstrate. Given a uvarint buffer with 10 bytes, it will result in `-9223372036854775808` (- 2^63). I tested the `ByteBuf` version too with the same result.

### Also...

Another minor issue I ran into while testing this out, I think something is wrong with the second part of the following check. I think it's supposed to be `index <= size` instead. Comparing the result (could be a big number) to the length of the byte array size doesn't make sense. It will only pass if the result has a value <=9.

https://github.com/libp2p/jvm-libp2p/blob/511c5a2f03c8401f658be15667326fd5e9dc8168/src/main/kotlin/io/libp2p/etc/types/ByteArrayExt.kt#L93

This function (the ByteArray version) is not yet used anywhere in the project, which is probably why it hasn't been caught.

jtraglia:
When reading over code, I noticed that it was possible to provide a 10-byte
uvarint (variable length integer) instead of the 9-byte maximum. The functions
did not properly catch it.

Also, fix what I believe is a typo in the ByteArray version. It checked the
wrong variable (result vs index) which would almost always result in a null
value. Might as well throw an exception there like the other implementation.

Fixes #225.

### DIFF
--- a/src/main/kotlin/io/libp2p/etc/types/ByteArrayExt.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/ByteArrayExt.kt
@@ -77,12 +77,9 @@ fun ByteArray.readUvarint(): Pair<Long, ByteArray>? {
 
     var index = 0
     var result: Long? = null
-    for (i in 0..9) {
+    for (i in 0..8) {
         val b = this.get(index++).toUByte().toShort()
         if (b < 0x80) {
-            if (i == 9 && b > 1) {
-                return null
-            }
             result = x or (b.toLong() shl s)
             break
         }
@@ -90,9 +87,10 @@ fun ByteArray.readUvarint(): Pair<Long, ByteArray>? {
         s += 7
     }
 
-    if (result != null && result <= size) {
+    // Check that we didn't break early.
+    if (result != null && index <= size) {
         return Pair(result, slice(IntRange(index, size - 1)).toByteArray())
     }
 
-    return null
+    throw IllegalStateException("uvarint too long")
 }
--- a/src/main/kotlin/io/libp2p/etc/types/ByteBufExt.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/ByteBufExt.kt
@@ -29,7 +29,7 @@ fun ByteBuf.readUvarint(): Long {
     var s = 0
 
     val originalReaderIndex = readerIndex()
-    for (i in 0..9) {
+    for (i in 0..8) {
         if (!this.isReadable) {
             // buffer contains just a fragment of uint
             readerIndex(originalReaderIndex)
@@ -37,9 +37,6 @@ fun ByteBuf.readUvarint(): Long {
         }
         val b = this.readUnsignedByte()
         if (b < 0x80) {
-            if (i == 9 && b > 1) {
-                throw IllegalStateException("Overflow reading uvarint")
-            }
             return x or (b.toLong() shl s)
         }
         x = x or (b.toLong() and 0x7f shl s)
--- a/src/main/kotlin/io/libp2p/etc/types/ByteBufExt.kt
+++ b/src/main/kotlin/io/libp2p/etc/types/ByteBufExt.kt
@@ -9,6 +9,10 @@ import java.lang.Integer.min
 fun ByteBuf.writeUvarint(value: Int): ByteBuf = writeUvarint(value.toLong())
 
 fun ByteBuf.writeUvarint(value: Long): ByteBuf {
+    if (value < 0) {
+        throw IllegalArgumentException("uvarint value must be positive")
+    }
+
     var v = value
     while (v >= 0x80) {
         this.writeByte((v or 0x80).toInt())
--- a/src/test/kotlin/io/libp2p/core/multiformats/MultiaddrTest.kt
+++ b/src/test/kotlin/io/libp2p/core/multiformats/MultiaddrTest.kt
@@ -191,7 +191,6 @@ class MultiaddrTest {
         fun invalidSerializations() = listOf(
             // Invalid var lengths
             Arguments.of(Protocol.DNS4.encoded + 1L.toVarInt()),
-            Arguments.of(Protocol.DNS4.encoded + (-1L).toVarInt()),
             Arguments.of(Protocol.DNS4.encoded + 10L.toVarInt() + ByteArray(9, { 0 })),
             Arguments.of(Protocol.DNS4.encoded + 65535L.toVarInt() + ByteArray(9, { 0 })),
             Arguments.of(Protocol.DNS4.encoded + Int.MAX_VALUE.toLong().toVarInt() + ByteArray(9, { 0 })),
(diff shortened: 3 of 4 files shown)


ajsutton:
Sorry for the delay on this one - I haven't forgotten it, just not an area I'm particularly familiar with so harder to review.

jtraglia:
No worries, I totally understand. There's no rush.

ajsutton (approved):
Ok, dug into the specs and this looks right to me - sorry it took so long. I've just added an additional test to confirm that we can successfully encode and decode `Long.MAX_VALUE` to ensure that the maximum length of 9 bytes is decoded.

jtraglia:
@ajsutton Just so you know, `testRoundTripMaximumValue` appears to be a duplicate of `testEncodeDecodeMaxLong`:

https://github.com/libp2p/jvm-libp2p/blob/5c3943d1f346d8a345e0a790b2ab773d22eb0e23/src/test/kotlin/io/libp2p/etc/types/UvarintTest.kt#L28-L33

https://github.com/libp2p/jvm-libp2p/blob/5c3943d1f346d8a345e0a790b2ab773d22eb0e23/src/test/kotlin/io/libp2p/etc/types/UvarintTest.kt#L61-L66

ajsutton:
Oh darn I didn't spot that. I'll have to clean that up. Thanks.

PR #227, opened 2022-02-28
Rename variable "interfaze" -> "interface"

jtraglia:
Simply rename all instances of this variable. Might have been a joke, but I think it's worth fixing.


ajsutton (approved):
LGTM. I suspect this was originally `interfaze`  because `interface` is a reserved word but with the `_` prefix it's fine either way.

PR #228, opened 2022-03-01
Fix various typos

jtraglia:
Two notable exceptions:

  (1) Changed "unrecognized" to "unrecognised" because there are two other
      instances of that word in the same file and this one stood out.

  (2) Added equal sign to a few gossip param error messages. Other params that
      accept zero for disabled values include "=" in the error message.


ajsutton (approved):
LGTM. Thanks for this - always good to get the details right.

Issue #223, opened 2022-02-25
Integrate detekt into build


jtraglia:
[Detekt](https://github.com/detekt/detekt) is a static code analysis tool for the Kotlin programming language. It operates on the abstract syntax tree provided by the Kotlin compiler, which is similar to how Google's [Error Prone](https://github.com/google/error-prone) tool works.

If this is something the team wants, I would be happy to work on this. I suspect it will some time to fix all of the minor issues and get the codebase to a compliant state. This could also be done incrementally, where each ruleset fix is its own PR.

PR #229, opened 2022-03-02
Add detekt with baseline

jtraglia:
Add detekt to the build system with all of the current findings suppressed via a baseline file. I will fix detekt findings and remove items from the baseline in subsequent PRs. Hoping to do one ruleset per PR.

Also, reorganize the plugin section in the gradle build file. There were differing ways of doing the same thing. Now everything is similar/sorted. The only important difference is the addition of `detekt`.

Related to #223.


ajsutton:
I'm not familiar with detekt and how useful vs noisy it can be.  @Nashatyrev would be really useful to get your thoughts on this one. Having something equivalent to errorprone would be good to flag likely problems but not create a lot of false positives, just got to make sure we get the right balance to avoid creating a lot of busywork suppressing warnings.

jtraglia:
Good point @ajsutton. By default, detekt checks for a lot more than errorprone, which isn't necessarily a good thing. There are many style and formatting rulesets that should probably be disabled (_e.g._ [MagicNumber](https://detekt.dev/style.html#magicnumber), [VariableNaming](https://detekt.dev/naming.html#variablenaming), and all of the [complexity](https://detekt.dev/complexity.html) rules). I am most interested in the potential bug rules, which are like errorprone:

* https://detekt.dev/potential-bugs.html

Nashatyrev:
@jtraglia thanks for your PR and sorry for delayed response 

Just went through the whole list and found just two potentially useful structural warnings in the ignore list:
- `EqualsWithHashCodeExist`: checks that both `equal` and `hashCode` are overridden. In our case this doesn't indicate error but potentially worth to double check such occurrences
-  `SwallowedException`: the same here: no actual errors in existing code, but potentially worth indicating 

The others to be honest are purely stylistical things which I would not enforce
I would suggest to exclude all types of warnings listed in your `baseline.xml` with the exception of two mentioned above. 
Does it make sense?

jtraglia:
Hey there. No worries. I have excluded the warnings (except for those two) in the baseline & all of the style/naming/complexity checks. Would you like for me to take a look at the `EqualsWithHashCodeExist` and `SwallowedException` findings to either fix or suppress?

Nashatyrev (approved):
LGTM 
Thanks for PR!
If you want you can address the remaining warnings as a separate PR and as a part of this one. It's on your choice :)

jtraglia:
I'll fix those in separate PR. Thanks!

Nashatyrev:
Ok, merging it then. Going to add new PR and would like to pick these changes up

PR #232, opened 2022-03-03
Add GossipParam tests and fix a few small issues

jtraglia:
* For gossipFactor, there was a range mismatch.
* For graylistThreshold, the message said "gossipThreshold" instead.
* The others checked >= or <= but the message said > or <.


jtraglia:
Before merging this, please look at #233. It is possible that the error messages were correct.

jtraglia:
Sorry this branch evolved into something else. Changing the name to reflect that.

jtraglia:
To be exhaustive, should I add negative checks to other values that should never be negative? For example all of the durations. It feels like it could be too much.

ajsutton (approved):
LGTM. I could see it being reasonable for graylistThreshold, publishThreshold and gossipThreashold to be equal but given the spec says less than that works even if it isn't always entirely consistent about it.

PR #235, opened 2022-03-25
Add listen/dial `preHandler` optional parameter

Nashatyrev:
This is more generic version of `DebugBuilder.beforeSecureHandler`: the raw channel pre-handler which could be specified per `listen()/dial()` call individually.

Add unit test for new `Transport` methods.
Also piggy tailing suppress unchecked warning for the test


tbenr (approved):
LGTM
pinging @ajsutton just for a double-check since I'm not 100% familiar with the code base

PR #236, opened 2022-03-25
Fix detekt's EqualsWithHashCodeExist warnings

jtraglia:
This PR fixes the [EqualsWithHashCodeExist](https://detekt.dev/potential-bugs.html#equalswithhashcodeexist) according to what detekt says should be done in cases of inheritance.

> All hash-based collections depend on objects meeting the equals-contract. Two equal objects must produce the same hashcode. **When inheriting equals or hashcode, override the inherited and call the super method for clarification.**

Also added a default `Key.hashCode()` function.

### DIFF
--- a/detekt/baseline.xml
+++ b/detekt/baseline.xml
@@ -2,18 +2,6 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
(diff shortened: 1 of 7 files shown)


Nashatyrev (approved):
LGTM

Nashatyrev:
@jtraglia Oh sorry, thought you were able to merge it :(

PR #239, opened 2022-05-09
Make cached sha256 of message data available

ajsutton:
Rather than caching the hashCode itself, caches the sha256 of the message content and makes it available via an API.  This enables applications using custom seen caches to leverage the same cached sha256 value for collision resistant hashing without having to recalculate the sha256.


rolfyone (approved):
LGTM

PR #244, opened 2022-07-04
Optimization/gossip peer scores

Nashatyrev:


Nashatyrev:
Forgot that score changes with time. It looks like more deep refactoring is needed for that. 
Closing this for now

PR #245, opened 2022-07-13
Optimize GossipRouter.mergeMessageParts()

Nashatyrev:
Currently the `GossipRouter.mergeMessageParts()` method is naively implemented and has ~ `O(n^2)` complexity. 
When testing with a lot of peers (> 1000) and all subnet subscriptions this method becomes a significant bottleneck. 

I've extracted the parts merge logic into a separate class and optimize message parts merge. The complexity of updated method should be `O(n)`


src/test/kotlin/io/libp2p/pubsub/gossip/GossipRpcPartsQueueTest.kt
```diff
@@ -0,0 +1,262 @@
...
+import pubsub.pb.Rpc
+import java.util.Collections
+import java.util.stream.Stream
+
+class GossipRpcPartsQueueTest {
+
+    class TestGossipQueue(params: GossipParams) : DefaultGossipRpcPartsQueue(params) {
+
+        fun shuffleParts() {
+            Collections.shuffle(parts)
```
ajsutton, line 21: IntelliJ would prefer this as:
```suggestion
            parts.shuffle()
```

ajsutton (approved):
LGTM. Great work - makes the code a lot more readable.

src/test/kotlin/io/libp2p/pubsub/gossip/GossipRpcPartsQueueTest.kt
Nashatyrev: Right. Fixed. 
Thought it creates a new shuffled instance

PR #246, opened 2022-07-13
Optimization: no need to instantiate SecureRandom on sign() calls

Nashatyrev:
Currently `Secp256k1PrivateKey.sign()` instantiates `SecureRandom` on every call. 
It's obsolete, as BouncyCastle uses it's own `SecureRandom` singleton when not supplied one.

### DIFF
--- a/src/main/kotlin/io/libp2p/crypto/keys/Secp256k1.kt
+++ b/src/main/kotlin/io/libp2p/crypto/keys/Secp256k1.kt
@@ -18,11 +18,7 @@ import io.libp2p.core.crypto.PrivKey
 import io.libp2p.core.crypto.PubKey
 import io.libp2p.core.crypto.sha256
 import io.libp2p.crypto.SECP_256K1_ALGORITHM
-import org.bouncycastle.asn1.ASN1InputStream
-import org.bouncycastle.asn1.ASN1Integer
-import org.bouncycastle.asn1.ASN1Primitive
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/libp2p/crypto/keys/Secp256k1.kt
```diff
@@ -18,19 +18,14 @@ import io.libp2p.core.crypto.PrivKey
 import io.libp2p.core.crypto.PubKey
 import io.libp2p.core.crypto.sha256
 import io.libp2p.crypto.SECP_256K1_ALGORITHM
-import org.bouncycastle.asn1.ASN1InputStream
-import org.bouncycastle.asn1.ASN1Integer
-import org.bouncycastle.asn1.ASN1Primitive
-import org.bouncycastle.asn1.ASN1Sequence
-import org.bouncycastle.asn1.DERSequenceGenerator
+import org.bouncycastle.asn1.*
```
ajsutton, line 21: Do we use `*` imports normally? I'm not overly fussed just checking.

Nashatyrev: Yep, it's kind of historically: 
```
kotlinter {
    disabledRules = arrayOf("no-wildcard-imports")
}
```

Tried to remove this line. The formatter was unable to fix this automatically. So leaving this 'as is' for now. Have no strong preference of this or that option

PR #248, opened 2022-07-14
Optimizate GossipScore IP colocation calculation

Nashatyrev:
The current algorithm of IP colocation score calculation has `O(n^2)` complexity and consumes significant CPU with lots of peer connections

This PR:
- Modifies the algorithm to `O(n)` complexity 
- Extracts the `PeerIP` class (which could be adjusted to IPv6 and colocation criteria later) 
- Pass more specific `Multiaddr` type for the `onConnected()` event


ajsutton (approved):
LGTM.

PR #249, opened 2022-07-18
Refactor GossipRouter: extract GossipRouterBuilder

Nashatyrev:
This is follow up to #247 

In the very initial version I've made an attempt to invent a new Kotlin specific pattern which combines both builder and its resulting class entities. It should in theory make the code smaller and supporting efforts cheaper. 

But it finally failed to be both a builder pattern and any kind of sustainable code. With `GossipRouter` becoming more complex class that approach is turning into error-prone spaghetti code. 

This PR extracts `GossipRouterBuilder` with making `GossipRouter` properties final (where appropriate), which looks more like a traditional pattern for Teku code and should be less error-prone and more sustainable 

Also:
- Move property declarations right after primary constructor (which may also declare properties) in changed classes
- Refactor `AbstractRouter.addPeerWithDebugHandler()` workaround 
- Remove pubsub interface `var` properties 
- Adjust tests

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -248,7 +248,7 @@ fun findProperty(s: String) = project.findProperty(s) as String?
 val compileKotlin: KotlinCompile by tasks
 compileKotlin.kotlinOptions {
     languageVersion = "1.6"
-    allWarningsAsErrors = false
+    allWarningsAsErrors = true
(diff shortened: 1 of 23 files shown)


tbenr (approved):
LGTM.

PR #251, opened 2022-08-05
Optimize Gossip peers <=> topics map

Nashatyrev:
With a large number of peers quadratic complexity of the `AbstractRouter.getTopicPeers()` method becomes a significant bottleneck. 

- Create the `MultiBiMap` class with `O(1)` access complexity 
- Use it in `AbstractRouter`


ajsutton (approved):
LGTM.

PR #252, opened 2022-08-05
Switch lib version to 'develop' in the develop branch

Nashatyrev:
Set the lib version to `develop` in the develop branch. 
This is for better local develop build testing. I.e. when you make `publishToMavenLocal` of a develop branch and need to test it with Teku of other project you may just use the `develop` maven dependency version


ajsutton (approved):
LGTM.

PR #253, opened 2022-08-15
Add JMH benchmark 

Nashatyrev:
Generally introduces the ability to add JMH benchmarks

For that: 
- extract `testFixtures` module from the `test` module. 
- add `jmh` source set and `jmh` build target (similar to the Teku gradle build) 
- make `jmh` module depending on `testFixtures`
- add a sample `GossipScoreBenchmark` benchmark. 
 
The benchmark is in Java since JMH doesn't work well with Kotlin. Tried the [kotlinx-benchmark](https://github.com/Kotlin/kotlinx-benchmark) suite, but unfortunately it doesn't work with existing JMH IDEA plugin


ajsutton (approved):
LGTM.

PR #261, opened 2022-12-19
Remove the interop tests

### LINKED ISSUE #260: test/external: Use rust-libp2p spec compliant `X25519Spec`
https://github.com/libp2p/jvm-libp2p/blob/1e6f17fbb216bcd000551dd413f4a8ec502b0f9f/src/test/external/rust/ping-client/src/main.rs#L13

We are deprecating `X25519`, please use `X25519Spec` instead.

See https://github.com/libp2p/rust-libp2p/pull/3227/.

ajsutton:
They weren't running in CI and haven't been maintained so the other language client no longer even compile cleanly. Given we're not running these tests and they were only basic ping, they aren't providing enough value to justify fixing and maintaining them.

fixes #260


Nashatyrev (approved):
Ok, let's remove them for now. Especially assuming they are not compiled any more. 
In case we want them back they are always in the git history 

I believe we should also remove `ServerInterOpTest` and `ClientInterOpTest` then

PR #264, opened 2023-02-10
Update dependencies and linting changes

### LINKED ISSUE #242: Netty 4.1.69 don't compile on android
api("io.netty:netty-all:4.1.69.Final") prevent libp2p from compiling my Android app.

Upgrading to 4.1.77 seem to fix the problem

After some experiences I also had to exclude log4j-core

StefanBratanov:
- Update versions of most libraries and gradle plugins
- Upgrade the gradle wrapper to 7.6
- Update `kotlin` version from 1.6.10 to 1.6.21
- Replace jmh initialization with a `me.champeau.jmh` gradle plugin
- Linting changes to files based on the `lintKotlin` task
- Change the `jvmTarget`, `sourceCompatibility` and `targetCompatibility` to Java 11 (align with the README.md saying Java 11 is needed.)
- Small `build.gradle` nits suggested by IntelliJ
- Change the two dependency badges in the README.md with one provided by cloudsmith

fixes #242


Nashatyrev:
All looks good to me except I would not upgrade Kotlin version unless it's strictly necessary
- Kotlin 1.8 is still marked as 'experimental' 
- Kotlin doesn't maintain strict backward compatibility. That may also break existing lib user's codebase 
- Kotlin is still not that stable as Java.

I also had some doubts regarding usage of `me.champeau.jmh` gradle plugin as we had some issues with it in Teku and finally replaced it with a custom code (https://github.com/ConsenSys/teku/pull/4817). But I tried running JMH from IDEA and from CLI and didn't noticed any problems. So we could apply this change for now and revert back in case if anything goes wrong.

StefanBratanov:
> All looks good to me except I would not upgrade Kotlin version unless it's strictly necessary
> 
> * Kotlin 1.8 is still marked as 'experimental'
> * Kotlin doesn't maintain strict backward compatibility. That may also break existing lib user's codebase
> * Kotlin is still not that stable as Java.
> 
> I also had some doubts regarding usage of `me.champeau.jmh` gradle plugin as we had some issues with it in Teku and finally replaced it with a custom code (https://github.com/ConsenSys/teku/pull/4817). But I tried running JMH from IDEA and from CLI and didn't noticed any problems. So we could apply this change for now and revert back in case if anything goes wrong.

Thanks Anton for the review. What is latest version of Kotlin which is not experimental or just reverting back to 1.6.10 makes more sense?

Nashatyrev:
> What is latest version of Kotlin which is not experimental or just reverting back to 1.6 makes more sense?

I would leave 1.6 unless there is a need for upgrade

StefanBratanov:
> > What is latest version of Kotlin which is not experimental or just reverting back to 1.6 makes more sense?
> 
> I would leave 1.6 unless there is a need for upgrade

Cool. Makes sense. I changed it to the latest version of 1.6. Had to downgrade couple of libraries to previous versions.

build.gradle.kts
```diff
@@ -40,7 +40,7 @@ dependencies {
     api("io.netty:netty-all:4.1.87.Final")
     api("com.google.protobuf:protobuf-java:3.21.12")
 
-    implementation(kotlin("stdlib"))
+    implementation(kotlin("stdlib-jdk8"))
```
Nashatyrev, line 43: I believe no need to revert this lib back to `stdlib-jdk8` as we switched to Java-11 target

Nashatyrev:
> Had to downgrade couple of libraries to previous versions.

Yep, that's the downside of not upgrade. Seems like no critical libs has to be downgraded though

build.gradle.kts
StefanBratanov: From what I read from Kotlin 1.8 [stdlib-jdk8 is merged into kotlin-stdlib](https://kotlinlang.org/docs/whatsnew18.html#updated-jvm-compilation-target) . Otherwise, it is preferred to use `stdlib-jdk8` to make use of the jdk8 extensions.

StefanBratanov:
Just force pushed a cleaner commit without a lot of linting changes (which applies only to the later versions of Kotlin and the kotlinter)

PR #198, opened 2021-09-18
Consolidate both Gradle projects into one

SalvatoreT:
Combine the "example" projects into the main project. I needed to move
the main library into it's own module so the other libraries could
exist. This allows a developer to open this project in IntelliJ and
build all of the projects easily.

There's still more clean-up to do such as consolidating the Kotlin
versions to one top level declaration, but I'm trying to make the
minimal amount of changes to get this to work.

Closes https://github.com/libp2p/jvm-libp2p/issues/153
Closes https://github.com/libp2p/jvm-libp2p/issues/152

PR #267, opened 2023-02-17
Convert to multi-module project

Nashatyrev:
The aim is to make this project multi-module to later add the gossip simulator module which now lives in tests in a separate branch. 
- extract the main module as a `libp2p` submodule 
- move specific dependencies and gradle plugins to the `libp2p` module
- remove unused dependencies
- add examples as submodules and adjust their code 
- android example is still out of the scope as I couldn't make it working in a reasonable time. Need to file an issue on this. 
- modules which want their artifacts to be published should define `mavenArtifactId` project property. Now just `libp2p` module is published under the old artifact id `jvm-libp2p-minimal`
- remove 'log4j-core' from 'libp2p' module: it's needed in runtime for tests or applications only
- extract dependency versions to the dedicated file

This PR was inspired by #198, thanks to @SalvatoreT


Nashatyrev:
@StefanBratanov had approved this PR

PR #269, opened 2023-02-21
Add Gossip simulator module 

Nashatyrev:
Add Gossip simulator module. 

- Extract `libp2p.testFixtures` `schedulers` package to a dedicated `:tools:schedulers` module 
- Add `:tools:simulator` module
- Simulation main classes are located in the package `io.libp2p.simulate.main`
- Any simulation could be run from the gradle command line (gradle `application` plugin applied in the module)
- Simulations are configured programmatically only
- Add simulator `README` and the reference from the main `README`


tools/simulator/README.md
```diff
@@ -0,0 +1,33 @@
...
+and change simulation configuration right in the Kotlin class 
+
+## Running simulation with gradle 
+
+Any main function could be run from CLI with gradle using the following syntax: 
+```shell
+> gradle :tools:simulator:run -PmainClass=<your.main.Class> [--args="you args"]
+```
+
+For example to run the sample simulation use the command below:
```
zilm13, line 25: Simple simulation?

zilm13 (approved):
LGTM

Do I understand correctly, that schedulers were not referenced from anywhere inside module before?

Nashatyrev:
> Do I understand correctly, that schedulers were not referenced from anywhere inside module before?

Schedulers were used by tests and JMH benchmark. Now also used by the simulator

tools/simulator/README.md
Nashatyrev: Sample simple simulation :)

Nashatyrev:
I've added a number of changes: reactoring and cleaning up the simulator code only. 
@zilm13 please let me know if I'm still good to go with these changes

PR #273, opened 2023-05-10
Add Yamux and TLS and fix multiaddr (WIP)

ianopolous:
This is now rebased off upstream, but it was more painful than I hoped because of the build system changes. It's currently not building, but I've moved all the new source files to the right places. If someone could give me a hand fixing the build file then I can work on any remaining issues.


ianopolous:
Split into smaller PRs

PR #275, opened 2023-05-15
Integrate Android example

### LINKED ISSUE #270: Add Android example as a gradle module
jvm-Libp2p was converted to multi-module gradle project and the examples were added as submodules except of [Android example](https://github.com/libp2p/jvm-libp2p/tree/develop/examples/android-chatter) 

Would be good to have it integrated as well.
Some requirements: 
- it should be skipped by gradle if no Android SDK is installed. I.e. it should work as before for those unaware of Android
- the module should be correctly imported to IntelliJ IDEA whenever Android SDK/Tools/Plugins installed or not

Nashatyrev:
Adds Android example to the gradle build to ensure jvm-Libp2p is interoperable with Android SDK: 
- Upgrade Gradle version to `8.0`
- Refactor Android example 
- Add 2 Android gradle plugins (without applying) to the root build (haven't managed to isolate them in Android example module)
- Add `android.useAndroidX` property to the root `gradle.properties` (for some reason defining this property in the Android submodule doesn't work)
- Add `libp2p.gradle.custom` per project property which indicates whether common gradle code for submodules should be skipped (set it up for Adnroid example)
- Move Maven repositories definitions to `settings.gradle`
- In the `settings.gradle` detect if Android SDK is installed and conditionally include `:examples:android-chatter` module
  - check if env var `ANDROID_HOME` is defined
  - or `sdk.dir` property is defined in `local.properties` file (created by AndroidStudio)
- Add `Setup Android SDK` GitHub action 

As a result: 
- When no Android SDK is installed: 
  - Gradle build works as before, Android example is not included into the build 
  - The project is imported into IntelliJ IDEA as before: Android example is not included as a module
- When Android SDK is installed and env var `ANDROID_HOME` is defined, Android example is build by the gradle 
- When the project is imported to AndroidStudio (`local.properties` file is automatically created), the Android example module is included and can be executed

Fix #270


StefanBratanov (approved):
LGTM

Issue #272, opened 2023-05-02
Upstream new protocols (Yamux, TLS, Quic)


ianopolous:
Hi we're hoping to upstream a bunch of new protocols we've implemented based on this. Specifically Yamux, TLS, multiaddr fixes and soon Quic. We also have Kademlia and bitswap in Java on top of this here: https://github.com/Peergos/nabu These are all tested against kubo for interop.

The only potential difficulties we see are
* the ~1 commit after we forked where this repo was refactored and a lot of stuff was moved around. 
* our removal of a bunch of dependencies like guava, log4j, apache commons
* our minimum Java 17 requirement (required for Ed25519 TLS handling)

Is there interest in up-streaming our work? What are your thoughts?

The fork is https://github.com/peergos/jvm-libp2p

mxinden:
Friendly ping @Nashatyrev.

Nashatyrev:
Hey, that sounds exciting!
However I believe that should be discussed more closely with the Teku team, since the changes could potentially impose the risk of regression bugs in the Teku networking layer.

In the meanwhile let me ask you some initial questions on the items you listed: 

> our removal of a bunch of dependencies like guava, log4j, apache commons

What's the reasons behind removing those dependencies?
Are you using other logging API? (I think I've seen Java Logging?) Don't you have plans to switch? 

> our minimum Java 17 requirement (required for Ed25519 TLS handling)

Teku is still on Java 11, so I'm not sure how we could handle this at the moment. As an option: postpone TLS module until Teku upgrade?

> the ~1 commit after we forked where this repo was refactored and a lot of stuff was moved around.

Are you referring to #267 ? I believe it shouldn't be too problematic to adopt, since there were mostly structural changes and the code was not affected mostly

ianopolous:
Thanks @Nashatyrev , yep that makes sense to discuss with Teku. 

We try to minimise dependencies for security and maintainability reasons mainly, but also compiled jar size. For logging in particular we've never needed log4j in various companies over the years and always manage with the built-in logging. 

I'd love to remove the Java 17 requirement too. It might be possible with some vendoring of bouncy castle components. Quic also depends on TLS though, so that would also have to be post-poned.  

Good to hear #267 shouldn't be difficult to rebase off.

Nashatyrev:
> our removal of a bunch of dependencies like guava, log4j, apache commons

- `commons` and `guava`: there are quite few classes used, so it probably makes sense to get rid of those deps 
[...]

PR #279, opened 2023-05-17
Migrate to slf4j

Nashatyrev:
`Slf4j` is a thinner and more commonly used logging abstraction library

- Replace `log4j-api` entries and dependencies with corresponding `slf4j` analogs 
- Add `log4j-slf4j2-impl` runtime dependencies for examples, tests and `simulator` module

Relates to #272


libp2p/src/main/java/io/libp2p/discovery/mdns/impl/DNSIncoming.java
```diff
@@ -27,14 +27,14 @@
  * @author Arthur van Hoff, Werner Randelshofer, Pierre Frisch, Daniel Bobbert
  */
 public final class DNSIncoming extends DNSMessage {
-    private static Logger logger                                = LogManager.getLogger(DNSIncoming.class.getName());
+    private static Logger logger                                = LoggerFactory.getLogger(DNSIncoming.class.getName());
```
lucassaldanha, line 30: Why so many spaces before the `=`? 😅

lucassaldanha (approved):
LGTM. However, there are some places where we have a bunch of empty spaces between variable name and assignment. Not sure if that was on purpose :)

libp2p/src/main/java/io/libp2p/discovery/mdns/impl/DNSIncoming.java
Nashatyrev: That was just copied sources 'as is'. If at some point we will add Java Spotless that would be fixed :)

PR #280, opened 2023-05-18
Support webtransport multiaddrs

kevodwyer:
as per line item "PR 2: multiaddr fixes"
referenced in: https://github.com/libp2p/jvm-libp2p/issues/272


libp2p/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -180,14 +186,21 @@ private val BASE58_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
 private val BASE58_STRINGIFIER: (Protocol, ByteArray) -> String = { _, bytes ->
     Base58.encode(bytes)
 }
+private val BASE64_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
```
Nashatyrev, line 189: Should we rather name it `MULTIBASE_PARSER` ?

libp2p/src/main/kotlin/io/libp2p/core/multiformats/Protocol.kt
```diff
@@ -180,14 +186,21 @@ private val BASE58_PARSER: (Protocol, String) -> ByteArray = { _, addr ->
...
+    Multibase.decode(addr)
+}
+private val BASE64_STRINGIFIER: (Protocol, ByteArray) -> String = { _, bytes ->
```
Nashatyrev, line 192: Should we rather name it `MULTIBASE_BASE64_STRINGIFIER` ?

Nashatyrev (approved):
LGTM 👍  Just minor naming suggestion

PR #283, opened 2023-05-24
Implement TLS security protocol including early muxer negotiation

ianopolous:
Early muxer negotiation is when the security protocol handshake is used to negotiate the muxer to avoid a round trip. In TLS this is done using ALPN. It is not implemented for Noise here.

This also includes a few small changes:
1) Remove the insecure Secio from builders so it can't be accidentally included
2) Upgrade to Java 17 which is required for Ed25519 TLS certs
3) Expose the event loop on Streams which we needed to implement p2p http proxies in Nabu

Part of https://github.com/libp2p/jvm-libp2p/pull/272

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
@@ -5,7 +5,11 @@ import io.libp2p.core.*
 import io.libp2p.core.crypto.PrivKey
 import io.libp2p.core.crypto.PubKey
 import io.libp2p.core.crypto.unmarshalPublicKey
+import io.libp2p.core.multistream.ProtocolBinding
 import io.libp2p.core.multistream.ProtocolDescriptor
+import io.libp2p.core.multistream.ProtocolId
+import io.libp2p.core.mux.NegotiatedStreamMuxer
+import io.libp2p.core.mux.StreamMuxer
 import io.libp2p.core.security.SecureChannel
 import io.libp2p.crypto.Libp2pCrypto
 import io.libp2p.crypto.keys.EcdsaPublicKey
@@ -52,13 +56,15 @@ import java.util.logging.Logger
 import javax.net.ssl.X509TrustManager
 
 private val log = Logger.getLogger(TlsSecureChannel::class.java.name)
-private val SetupHandlerName = "TlsSetup"
+
+const val NoEarlyMuxerNegotiationEntry = "libp2p"
+const val SetupHandlerName = "TlsSetup"
 val certificatePrefix = "libp2p-tls-handshake:".encodeToByteArray()
 
-class TlsSecureChannel(private val localKey: PrivKey, private val muxerIds: List<String>, private val certAlgorithm: String) :
+class TlsSecureChannel(private val localKey: PrivKey, private val muxers: List<StreamMuxer>, private val certAlgorithm: String) :
     SecureChannel {
 
-    constructor(localKey: PrivKey, muxerIds: List<String>) : this(localKey, muxerIds, "Ed25519") {}
+    constructor(localKey: PrivKey, muxerIds: List<StreamMuxer>) : this(localKey, muxerIds, "Ed25519") {}
 
     companion object {
         const val announce = "/tls/1.0.0"
@@ -79,7 +85,7 @@ class TlsSecureChannel(private val localKey: PrivKey, private val muxerIds: List
         selectedProtocol: String
     ): CompletableFuture<SecureChannel.Session> {
         val handshakeComplete = CompletableFuture<SecureChannel.Session>()
-        ch.pushHandler(SetupHandlerName, ChannelSetup(localKey, muxerIds, certAlgorithm, ch, handshakeComplete))
+        ch.pushHandler(SetupHandlerName, ChannelSetup(localKey, muxers, certAlgorithm, ch, handshakeComplete))
         return handshakeComplete
     }
 }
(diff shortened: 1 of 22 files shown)


ianopolous:
I've just noticed the old yamux changes got in there somehow. This was already my second attempt at cherry-picking onto v1.0.0.

Nashatyrev:
@ianopolous thanks for another contribution 👍 

Some questions here: 
- is that PR possible without switching to Java 17? We are potentially considering upgrading the lib to Java 17 at some point, however if possible I would prefer make that transition in a separate PR. 
  - Maybe BouncyCastle Ed25519 implementation could be used instead of JDK?
  - Is it possible to make Ed25519 support pluggable? 
  - As the last resort: could Ed25519 support be omitted for now?
- Could you please explain usecases for the `eventLoop()`? We were trying originally to keep Netty API exposure as minimal as possible. Exposing the Netty `EventLoop` (which is not quite a basic api) via generic `Stream` looks slightly painful for me. 

I'm planning to revisit early muxer handshake concept and probably make a separate PR for introducing it

ianopolous:
I think I'm most of the way to removing the need for java 17. I've pushed it to a new branch: https://github.com/Peergos/jvm-libp2p/tree/feat/upstream-tls-java-11

However I can't get gradle to find the new bouncy castle jar (which is in maven central like the rest). Any chance you could have a look @Nashatyrev ?

Nashatyrev:
> However I can't get gradle to find the new bouncy castle jar (which is in maven central like the rest). 

Wouldn't that help? 

```patch
Index: versions.gradle
IDEA additional info:
<+>UTF-8
diff --git a/versions.gradle b/versions.gradle
--- a/versions.gradle	(revision d4eb98c8bd4d6ae86eca228c35682b2da9edf9b7)
+++ b/versions.gradle	(date 1685016596704)
@@ -40,6 +40,7 @@
         dependencySet(group: "org.bouncycastle", version: "1.70") {
             entry 'bcprov-jdk15on'
             entry 'bcpkix-jdk15on'
+            entry 'bctls-jdk15on'
         }
     }
 }
```

Nashatyrev:
Do you mean it doesn't work with TLS security, but works with others? 
Or did it stop working after switching to BC TLS implementation?

ianopolous:
The test is specifically for TLS (If I switch to Noise it is fine). It is working in this branch on java 17, but not in the java 11 branch with BC.

ianopolous:
I think https://github.com/Peergos/jvm-libp2p/tree/feat/upstream-tls-java-11 is quite close. It works fine Java to Java using Ed25519 certs. Talking to Kubo (which uses ECDSA over P-256) the TLS handshake completes, but a subsequent protocol select times out.

Nashatyrev:
@ianopolous 
You may want to try adding debug wire loggers:
https://github.com/libp2p/jvm-libp2p/blob/89ba1c281a490eb036f974fe51ca3fc0f936ff00/libp2p/src/test/kotlin/io/libp2p/core/HostTest.kt#L138-L142
Just use `addLogger()` instead of `addHandler()`

ianopolous:
@Nashatyrev Is that possible from the Java Host test?

ianopolous:
The plot thickens. Running HostTestJava to Kubo works with {Noise + (mplex or yamux)}, but fails with TLS

However, if I disable early muxer negotiation, then TLS works! So it's possible we broke early muxer with some of the upstream changes??

ianopolous:
@Nashatyrev Ok. We're now good with Java 11.

The use for the EventLoop is a p2p http proxy. It accepts incoming libp2p connections and proxies it to a fixed http endpoint or http handler. We needed access to the eventloop to add the new http handlers when receiving a new connection: https://github.com/Peergos/nabu/blob/master/src/main/java/org/peergos/protocol/http/HttpProtocol.java#L103

Nashatyrev:
> @Nashatyrev Ok. We're now good with Java 11.

Great news! Thanks!
Sorry I'm still on the way with the early muxer negotiation integration

> The use for the EventLoop is a p2p http proxy. It accepts incoming libp2p connections and proxies it to a fixed http endpoint or http handler. We needed access to the eventloop to add the new http handlers when receiving a new connection: https://github.com/Peergos/nabu/blob/master/src/main/java/org/peergos/protocol/http/HttpProtocol.java#L103

Sorry, I didn't dive deeply into the usage case above.
Anyway I would suggest to discuss that `eventLoop` question in a separate PR/Issue

ianopolous:
@Nashatyrev I've removed the EventLoop stuff from this PR. Are you planning on PRing your early muxer stuff into this branch?

libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
```diff
@@ -0,0 +1,315 @@
...
+        .ciphers(listOf("TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256"))
+        .clientAuth(ClientAuth.REQUIRE)
+        .trustManager(Libp2pTrustManager(expectedRemotePeer))
+        .sslContextProvider(BouncyCastleJsseProvider())
+        .applicationProtocolConfig(
+            ApplicationProtocolConfig(
+                ApplicationProtocolConfig.Protocol.ALPN,
+                ApplicationProtocolConfig.SelectorFailureBehavior.FATAL_ALERT,
+                ApplicationProtocolConfig.SelectedListenerFailureBehavior.FATAL_ALERT,
+                muxerIds.plus("libp2p") // early muxer negotiation
```
Nashatyrev, line 114: Why do we need adding `"libp2p"` here? 
Just doing some minor refactoring and curious if it would be the same effect if we don't add this value?

ianopolous: This is for compatibility with clients that do not support early muxer negotiation (they just send "libp2p"). So definitely necessary.

Nashatyrev (approved):
Basically LGTM. Just please consider some minor refactor here: https://github.com/Peergos/jvm-libp2p/pull/9

I'm still slightly concerned with `SecureChannel.Session.earlyMuxer` property. It has a transient nature and is not intended to be exposed in the API. But I'm planning to address it later in a separate PR to not block further upstream work

PR #290, opened 2023-05-31
Expose eventloop on Stream

ianopolous:
The use for the EventLoop is a p2p http proxy. It accepts incoming libp2p connections and proxies it to a fixed http endpoint or http handler. We needed access to the eventloop to add the new http handlers when receiving a new connection: https://github.com/Peergos/nabu/blob/master/src/main/java/org/peergos/protocol/http/HttpProtocol.java#L103

It is basically HTTP over libp2p. The entire implementation is in the single file linked above. We explain the use case a bit more in the "Decentralization" section of https://peergos.org/posts/dev-update

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/core/Stream.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/core/Stream.kt
@@ -2,6 +2,7 @@ package io.libp2p.core
 
 import io.libp2p.protocol.ProtocolMessageHandler
 import io.libp2p.protocol.ProtocolMessageHandlerAdapter
+import io.netty.channel.EventLoop
 import java.util.concurrent.CompletableFuture
 
 /**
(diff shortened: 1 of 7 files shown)


Nashatyrev:
```java
    public static void proxyRequest(Stream stream,
                                    HttpRequest msg,
                                    SocketAddress proxyTarget,
                                    Consumer<HttpObject> replyHandler) {
        Bootstrap b = new Bootstrap();
        b.group(stream.eventLoop())
                .channel(NioSocketChannel.class)
                .handler(new LoggingHandler(LogLevel.TRACE));

        ChannelFuture fut = b.connect(proxyTarget);
        Channel ch = fut.channel();
        ch.pipeline().addLast(new HttpRequestEncoder());
        ch.pipeline().addLast(new HttpResponseDecoder());
        ch.pipeline().addLast(new ResponseWriter(replyHandler));

        fut.addListener(x -> ch.writeAndFlush(msg));
    }
```
I'm not sure why do you need `stream`'s eventLoop to create an independent connection?
Why not create a new `NioEventLoopGroup` or create and cache a single instance if it's performance critical?

ianopolous:
I wanted to reuse the same thread. We were originally using a new pool, but this was 1000X faster. However I've just benchmarked using a static instance and that doesn't seem any slower than the same thread. So I think you're right, we don't need this.

Nashatyrev:
Sorry, accidentally posted incomplete comment. Some edit updates to previous comment

Nashatyrev:
Was part of #272

Issue #291, opened 2023-05-31
Yamux memory leak


ianopolous:
I'm seeing this in Nabu in a bitswap unit test:

May 31, 2023 8:02:11 PM io.netty.util.ResourceLeakDetector reportTracedLeak
SEVERE: LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:403)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:179)
	io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:132)
	io.netty.handler.codec.MessageToByteEncoder.allocateBuffer(MessageToByteEncoder.java:140)
	io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:105)
	io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:881)
	io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:863)
	io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:968)
	io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:856)
	io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:113)
	io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:881)
	io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:863)
	io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:968)
	io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:856)
	io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:113)
	io.netty.handler.codec.MessageToMessageCodec.write(MessageToMessageCodec.java:116)
[...]

PR #292, opened 2023-06-02
Fix yamux remote open

ianopolous:
Bitswap and identify from kubo, on a connection we initiated, was failing because we were not writing the /multistream/1.0.0 because the send window had not been initialised yet for a remote opened stream in yamux.

We should also tolerate window updates which close a stream and have 0 size (rather than error).

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
@@ -75,6 +75,7 @@ open class YamuxHandler(
         when (msg.flags) {
             YamuxFlags.SYN -> {
                 // ACK the new stream
+                onStreamCreate(msg.id) // sometimes writes can happen before onRemoteCreated is called
                 onRemoteOpen(msg.id)
                 ctx.writeAndFlush(YamuxFrame(msg.id, YamuxType.WINDOW_UPDATE, YamuxFlags.ACK, 0))
             }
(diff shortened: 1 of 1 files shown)


ianopolous:
Also fixes https://github.com/libp2p/jvm-libp2p/issues/291

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -75,6 +75,7 @@ open class YamuxHandler(
         when (msg.flags) {
             YamuxFlags.SYN -> {
                 // ACK the new stream
+                onStreamCreate(msg.id) // sometimes writes can happen before onRemoteCreated is called
                 onRemoteOpen(msg.id)
```
Nashatyrev, line 79: That would result to `onStreamCreate()` is called twice. 
What's the intention of calling `onStreamCreate()` explicitly here?

Nashatyrev (changes requested):
Would be also really nice if you could add a unit test on any regression found and fixed 👍

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
ianopolous: The send window needs to be initialized before onRemoteOpen, otherwise it errors and fails to do the first write.

Nashatyrev:
@ianopolous please check https://github.com/Peergos/jvm-libp2p/pull/10

Nashatyrev (approved):
LGTM when merged https://github.com/Peergos/jvm-libp2p/pull/10

PR #293, opened 2023-06-05
Fix ecdsa cert usage in TLS which bouncycastle broke

ianopolous:
Also update unit test to cover this

Edit: Including another fix for early muxer negotiation via ALPN which wasn't actually using the ALPN result. 

Added a helper method to construct ECDSA based instances of TLS.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
@@ -145,7 +145,7 @@ fun buildTlsHandler(
                     mux.protocolDescriptor.protocolMatcher.matches(nextProtocol)
                 }
                 .map { mux ->
-                    NegotiatedStreamMuxer(mux, nextProtocol) 
+                    NegotiatedStreamMuxer(mux, nextProtocol)
                 }
                 .firstOrNull()
             handshakeComplete.complete(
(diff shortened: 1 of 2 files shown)


libp2p/src/main/kotlin/io/libp2p/security/tls/TLSSecureChannel.kt
```diff
@@ -132,8 +139,15 @@ fun buildTlsHandler(
...
-            val selectedMuxer = muxers.findBestMatch(negotiatedProtocols)
+            val nextProtocol = handler.applicationProtocol()
+            val selectedMuxer = muxers.stream()
+                .flatMap { mux ->
+                    mux.protocolDescriptor.announceProtocols.stream()
+                        .filter({ it.equals(nextProtocol) })
+                        .map { NegotiatedStreamMuxer(mux, it) }
+                }
+                .findFirst()
+                .orElse(null)
```
Nashatyrev, line 150:
```suggestion
            val selectedMuxer = muxers
                .filter { mux ->
                    mux.protocolDescriptor.protocolMatcher.matches(nextProtocol)
                }
                .map { mux ->
                    NegotiatedStreamMuxer(mux, nextProtocol) 
                }
                .firstOrNull()
```
It's better use `protocolDescriptor.protocolMatcher` instead of strict `ProtocolId` equals as it may potentially may differ (not yet the case if I remember correctly)

+ also use Kotlin collections instead of Java streams

Nashatyrev (approved):
LGTM. 
Just a small change request

PR #295, opened 2023-07-11
Fix yamux handling of writes bigger than the window size

ianopolous:
Also fix inefficient window size handling.

I tried to test this with ping, but of course ping doesn't handle messages sent in multiple parts. But we have tested this downstream.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
@@ -43,9 +43,10 @@ open class YamuxHandler(
                 val buf = buffered.first()
                 val readableBytes = buf.readableBytes()
                 if (readableBytes + written < sendWindow.get()) {
-                    buffered.removeFirst()
                     sendBlocks(ctx, buf, sendWindow, id)
                     written += readableBytes
+                    buf.release()
+                    buffered.removeFirst()
                 } else {
                     // partial write to fit within window
                     val toRead = sendWindow.get() - written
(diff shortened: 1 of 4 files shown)


mxinden:
In case it is of some use, in rust-yamux [we split data frames at `16* 1024` bytes](https://github.com/libp2p/rust-yamux/blob/dcff3d5bf454c8bf97e74be3a46393844cee0939/yamux/src/lib.rs#L50-L64).

> By limitting the data frame payload size one gains the following
> beneftis:
> 
> 1. Reduce delays sending time-sensitive frames, e.g. window updates.
> 
> 2. Minimize head-of-line blocking across streams.
> 
> 3. Enable better interleaving of send and receive operations, as each is
> carried out atomically instead of concurrently with its respective
> counterpart.
> 
> Limiting the frame size to 16KiB does not introduce a large overhead. A
> Yamux frame header is 12 bytes large, thus this change introduces an
> overhead of ~0.07%.

See https://github.com/libp2p/rust-yamux/pull/111 for details.

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -34,19 +34,25 @@ open class YamuxHandler(
         private val buffered = ArrayDeque<ByteBuf>()
 
         fun add(data: ByteBuf) {
-            buffered.add(data)
+            buffered.add(data.retain())
```
Nashatyrev, line 37: Please see the update on this `retain()`: https://github.com/Peergos/jvm-libp2p/pull/12

Nashatyrev (approved):
LGTM

PR #298, opened 2023-08-15
Add Yamux specific unit tests

### LINKED ISSUE #289: Add Yamux specific unit tests
This is the follow up to PR #285 where a number of generic muxer unit tests were added
However Yamux has a number of specific feature which are to be tested yet.
`YamuxHandlerTest` seems like a good starting point for adding new unit tests

StefanBratanov:
fixes #289


libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxFrame.kt
```diff
@@ -8,16 +8,18 @@ import io.netty.buffer.Unpooled
...
 /**
  * Contains the fields that comprise a yamux frame.
- * @param streamId the ID of the stream.
+ * @param id the ID of the stream.
  * @param flag the flag value for this frame.
+ * @param length the length field for this frame.
  * @param data the data segment.
  */
-class YamuxFrame(val id: MuxId, val type: Int, val flags: Int, val lenData: Long, val data: ByteBuf? = null) :
+class YamuxFrame(val id: MuxId, val type: Int, val flag: Int, val length: Long, val data: ByteBuf? = null) :
```
Nashatyrev, line 16: I believe the field is still `flags`. Though I'm not sure there are use cases when more than one flag is set. 
https://github.com/hashicorp/yamux/blob/master/spec.md#framing

StefanBratanov: You are right. Fixed it.

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxFrameCodec.kt
```diff
@@ -42,32 +42,44 @@ class YamuxFrameCodec(
...
         while (msg.isReadable) {
-            if (msg.readableBytes() < 12)
+            if (msg.readableBytes() < 12) {
                 return
+            }
             val readerIndex = msg.readerIndex()
             msg.readByte(); // version always 0
             val type = msg.readUnsignedByte()
-            val flags = msg.readUnsignedShort()
+            val flag = msg.readUnsignedShort()
```
Nashatyrev, line 51: nit: should that be named 'flags' as well

PR #299, opened 2023-08-15
Add Host.getProtocols() and fix addProtocolHandler

ianopolous:
addProtocolHandler wasn't working because of a CopyOnWriteArrayList

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/multistream/MultistreamImpl.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/multistream/MultistreamImpl.kt
@@ -15,8 +15,7 @@ class MultistreamImpl<TController>(
     val negotiationTimeLimit: Duration = DEFAULT_NEGOTIATION_TIME_LIMIT
 ) : Multistream<TController> {
 
-    override val bindings: MutableList<ProtocolBinding<TController>> =
-        CopyOnWriteArrayList(initList)
+    override val bindings: List<ProtocolBinding<TController>> = initList
 
     override fun initChannel(ch: P2PChannel): CompletableFuture<TController> {
         return with(ch) {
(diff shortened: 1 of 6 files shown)


libp2p/src/main/kotlin/io/libp2p/multistream/MultistreamImpl.kt
```diff
@@ -6,7 +6,6 @@ import io.libp2p.core.multistream.Multistream
 import io.libp2p.core.multistream.ProtocolBinding
 import java.time.Duration
 import java.util.concurrent.CompletableFuture
-import java.util.concurrent.CopyOnWriteArrayList
 
 class MultistreamImpl<TController>(
```
StefanBratanov, line 10: For simplicity can just add the `bindings` to the constructor
```kotlin
class MultistreamImpl<TController>(
    override val bindings: MutableList<ProtocolBinding<TController>>,
    val preHandler: P2PChannelHandler<*>? = null,
    val postHandler: P2PChannelHandler<*>? = null,
    val negotiationTimeLimit: Duration = DEFAULT_NEGOTIATION_TIME_LIMIT
)
```
and remove
`override val bindings: List<ProtocolBinding<TController>> = initList`

Nashatyrev: Agree, it's not that right to pass a `List` (which is implicitly immutable) which is mutated behind the scene and is expected to have effect

ianopolous: Sure, I can make that change, though it ripples through the classes a bit.

ianopolous: Actually I tried to do this and it goes down a long rabbit hole that I couldn't get to work.

Nashatyrev: Actually I think I have already come up with some more or less acceptable change. I'll try to polish it tomorrow

Nashatyrev: Well, I would suggest to merge that variant with a couple of updates: https://github.com/Peergos/jvm-libp2p/pull/13

Nashatyrev (approved):
LGTM

However it looks like the design needs to be reviewed

PR #306, opened 2023-08-21
Disable publishing of test fixtures variants

StefanBratanov:
https://docs.gradle.org/current/userguide/java_testing.html#ex-disable-publishing-of-test-fixtures-variants


Nashatyrev (approved):
LGTM

Issue #311, opened 2023-08-23
io.libp2p.core.Libp2pException: Overflowed send buffer for connection (Yamux)


StefanBratanov:
Testing libp2p in Teku and specifically Yamux. Getting errors such as below. Also attached the whole log file: [teku-yamux.log](https://github.com/libp2p/jvm-libp2p/files/12417262/teku-yamux.log)

```java
2023-08-23T07:51:34,235 ERROR - Failed to process blocks by range request
 java.util.concurrent.CompletionException: io.libp2p.core.Libp2pException: Overflowed send buffer for connection
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:781)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at tech.pegasys.teku.networking.p2p.libp2p.rpc.LibP2PRpcStream.lambda$toSafeFuture$1(LibP2PRpcStream.java:72)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:590)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:557)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:492)
	at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:636)
	at io.netty.util.concurrent.DefaultPromise.setFailure0(DefaultPromise.java:629)
	at io.netty.util.concurrent.DefaultPromise.tryFailure(DefaultPromise.java:118)
	at io.netty.util.internal.PromiseNotificationUtil.tryFailure(PromiseNotificationUtil.java:64)
	at io.netty.channel.ChannelOutboundBuffer.safeFail(ChannelOutboundBuffer.java:733)
	at io.netty.channel.ChannelOutboundBuffer.remove0(ChannelOutboundBuffer.java:318)
	at io.netty.channel.ChannelOutboundBuffer.remove(ChannelOutboundBuffer.java:298)
[...]

PR #314, opened 2023-09-08
[Yamux] Limit the send buffer size per stream

StefanBratanov:
Before we were limiting the sending buffer for the whole connection. I had a look in rust implementation and it looks like their implementation is limiting it per stream https://github.com/libp2p/rust-yamux/blob/master/yamux/src/lib.rs .This change could help with https://github.com/libp2p/jvm-libp2p/issues/311

- Changed currently configured MAX_BUFFER_SIZE to be per stream
- Renamed `sendBlocks` to `sendFrames`


StefanBratanov:
Closing this. Will increase the buffer size per connection instead.

PR #317, opened 2023-09-12
[Yamux] Increase write buffer size and make it configurable

StefanBratanov:
- Make `maxBufferedConnectionWrites` configurable and default it to 10MiB
- Release and clear  `buffer` when overflowed.


Nashatyrev (approved):
LGTM

Would be nice as well to add a test checking the parameter is respected

Nashatyrev (approved):
LGTM

PR #319, opened 2023-09-13
[Yamux] Send whole data if window size is > 0

StefanBratanov:
There were some inconsistencies with `sendFrames` method and the send buffer `flush` approach to sending data.

- `sendFrames` was checking if window size is <= 0 and adding to buffer if so and afterwards was slicing the data based on the window size and sending across. So essentially if window size is 10 and data is 100 bytes, it will send across 10 frames of 10 bytes.
- `flush` was respecting the window size and was sending partial data if the length of the data is less than the window size.

I would imagine a better approach is if window size is > 0, just send the whole data respecting the `maxFrameDataLength`. 

- Made `flush` not use `sendFrames` since the data is already limited by `maxFrameDataLength` so no slicing needed
- Renamed `sendFrames` to `sendData`
- Changed `sendData` to respect the window size and add to buffer there instead of `onChildWrite`
- Modified tests accordingly
- Release message when `onChildWrite` and no window size is found.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/core/mux/StreamMuxerProtocol.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/core/mux/StreamMuxerProtocol.kt
@@ -22,7 +22,7 @@ fun interface StreamMuxerProtocol {
         }
 
         /**
-         * @param maxBufferedConnectionWrites the maximum amount of bytes in the write buffer per connection before termination
+         * @param maxBufferedConnectionWrites the maximum amount of bytes in the write buffer per connection
          */
         @JvmStatic
         @JvmOverloads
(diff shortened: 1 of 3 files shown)


StefanBratanov:
> There were some inconsistencies with `sendFrames` method and the send buffe `flush` approach to sending data.
> 
> * `sendFrames` was checking if window size is <= 0 and adding to buffer if so and afterwards was slicing the data based on the window size and sending across. So essentially if window size is 10 and data is 100 bytes, it will send across 10 frames of 10 bytes.
> * `flush` was respecting the window size and was sending partial data if the length of the data is less than the window size.
> 
> I would imagine a better approach is if window size is > 0, just send the whole data respecting the `maxFrameDataLength`.
> 
> * Made `flush` not use `sendFrames` since the data is already limited by `maxFrameDataLength` so no further slicing needed
> * Renamed `sendFrames` to `sendData`
> * Changed `sendData` to respect the window size and add to buffer there instead of `onChildWrite`
> * Modified tests accordingly

Nashatyrev (approved):
LGTM

PR #320, opened 2023-09-13
[Yamux] Send RST when write buffer has overflowed

StefanBratanov:
- Send a reset frame when a write buffer has overflown. 
- Ensure we remove entry from `sendWindowSizes` to stop further writes on both FIN and RST
- Close buffer when RST and on `onChildClosed`
- Flush and close buffer on FIN


Nashatyrev (approved):
LGTM

PR #321, opened 2023-09-13
[Yamux] Clear caches when connection is closed

StefanBratanov:
Clear sendBuffers and windowSizes when connection is closed (`channelUnregistered` Netty callback)


Nashatyrev (approved):
LGTM

PR #318, opened 2023-09-12
[Yamux] Revert merging send and receive windows maps

StefanBratanov:
Based on https://github.com/libp2p/jvm-libp2p/pull/312#issuecomment-1714199823

PR #323, opened 2023-09-14
[Yamux] Refactor window sizes retrieval

StefanBratanov:
Revert changes in https://github.com/libp2p/jvm-libp2p/pull/318

Since it is a duplex stream and there is no notion of inbound/outbound and there is no possibility of 2 `MuxId` per stream, there is no need for a separate map.

- Created `WindowSize` data class to track both send and receive for a `MuxId`


Nashatyrev (approved):
LGTM

PR #326, opened 2023-09-15
Refactor YamuxHandler

Nashatyrev:
- Create `YamuxStreamHandler` inner class
- incapsulate all stream related operations there
- Add Muxer specific exceptions
- Adopt tests


StefanBratanov (approved):
LGTM

PR #328, opened 2023-09-15
Refactor YamuxHandler.SendBuffer

Nashatyrev:
* Introduce ByteBufQueue
* Add ByteBufQueue tests
* Writing exec path is always through fill/drain buffer
* Adopt/fix existing tests
* Add new test checking correct handling of negative sendWindowSize

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/util/netty/ByteBufQueue.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/util/netty/ByteBufQueue.kt
@@ -1,0 +1,44 @@
+package io.libp2p.etc.util.netty
+
+import io.netty.buffer.ByteBuf
+import io.netty.buffer.Unpooled
+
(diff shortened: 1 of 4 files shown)


StefanBratanov (approved):
LGTM

libp2p/src/main/kotlin/io/libp2p/etc/util/netty/ByteBufQueue.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        val partialBufferSlice =
+            when {
+                data.isEmpty() -> null
+                size == maxLength -> null
+                else -> data.first()
+            }
+                ?.let { buf ->
+                    val remainingBytes = maxLength - size
+                    buf.readSlice(remainingBytes).retain()
```
StefanBratanov, line 32: I think can use `readRetainedSlice()` instead. At least docs state it could be better.

PR #330, opened 2023-09-16
Fix the case when a stream is closed while still having buffered data for write

Nashatyrev:
When a Stream is locally closed for writing but there is some data buffered for writing still, the `YamuxHandler` shouldn't immediately close the Stream for writing (by issuing `FIN` frame) and discarding the write buffer. 

It should rather mark the Stream as closed for writing and wait for the suitable window update(s) to send the remaining data and only then issue a `FIN` frame.


StefanBratanov (approved):
LGTM

PR #332, opened 2023-09-26
[Yamux] Don't send frame if send buffer is not empty

StefanBratanov:
There could be a race condition case where a frame is being sent before the send buffer has been flushed, which would result in unordered frames. Very hard to replicate in actual test, so wrote a generic test to test order of messages when send buffer is used.


Nashatyrev (approved):
LGTM

PR #334, opened 2023-09-28
Yamux is broken

ianopolous:
@Nashatyrev There are 2 things in this PR

1) Don't try and dial DNSADDR addresses in TCP transport
These should mainly be used for bootstrap nodes with a DNS TXT record lookup to expand to list of nodes (which we've implemented downstream for kademlia). This removes a lot of log spam from failed dial attempts in Kademlia.

2) We tried updating downstream to include the recent yamux refactoring but it broke sending large blobs. We have a downstream unit test demonstrating it, but it relies on the bitswap protocol. Here I've implemented a new Blob protocol just for testing which sends a length encoded byte array and echos it back. This works for sizes < 1MiB-4, but hangs and times out for larger sizes.


Nashatyrev:
Yes, thanks for the unit test 👍
I have added it to the refactored Yamux branch and it passes here: https://github.com/libp2p/jvm-libp2p/pull/328
You may want to address `DNSADDR` separately, but it just make sense to wait for https://github.com/libp2p/jvm-libp2p/pull/335 merged to avoid that mess with the `gradlew.bat`

ianopolous:
@Nashatyrev Excellent. Any idea what might have broken/fixed it?

I wait, and redo the DNSADDR in a new PR later.

Nashatyrev:
Added `largeBlolb` test as a separate PR here #337

Nashatyrev:
> I will wait, and redo the DNSADDR in a new PR later.

Then I'm closing this PR

PR #338, opened 2023-10-10
Refactor Yamux flags

Nashatyrev:
(based atop of #330)
- Convert `YamuxType` to enum: [e088393](https://github.com/libp2p/jvm-libp2p/pull/338/commits/e0883934d6cc66683e3ee649866e8008ae0d328b)
- Refactor `YamuxFlags`: convert them to `Set` of `enum` values. While the Yamux spec doesn't seem to allow a flags bitmask with more than _one_ flag set it looks better to treat that explicitly as a 'set of flags': [e28efd5](https://github.com/libp2p/jvm-libp2p/pull/338/commits/e28efd557fdf8e9ad685c9970e4188ea8a3ca0bf)


libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxFrame.kt
```diff
@@ -9,17 +8,15 @@ import io.netty.buffer.Unpooled
 /**
  * Contains the fields that comprise a yamux frame.
  * @param id the ID of the stream.
- * @param flags the flags value for this frame.
+ * @param flag the flag for this frame.
```
StefanBratanov, line 11: should be flags right?

StefanBratanov (approved):
LGTM with a nit

PR #340, opened 2023-10-24
[Yamux] Allow max ACK backlog of 256 streams

StefanBratanov:
https://github.com/libp2p/specs/blob/master/yamux/README.md#ack-backlog--backpressure

Implements `SHOULD at most allow an ACK backlog of 256 streams.`

- Added new parameter `ackBacklogLimit` when initializing Yamux
- Introduced notion of `acknowledged` when initializing yamux stream
- throw and don't create new stream if more than 256 streams are not acknowledged

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
@@ -92,9 +92,7 @@ open class YamuxHandler(
                 }
 
                 YamuxFlag.ACK in msg.flags -> {
-                    if (outbound) {
-                        acknowledged.set(true)
-                    }
+                    acknowledgeOutboundStreamIfNeeded()
                 }
 
                 YamuxFlag.FIN in msg.flags -> onRemoteDisconnect(msg.id)
@@ -104,6 +102,12 @@ open class YamuxHandler(
 
         private fun acknowledgeInboundStreamIfNeeded() {
             if (!outbound) {
+                acknowledged.set(true)
+            }
+        }
+
+        private fun acknowledgeOutboundStreamIfNeeded() {
+            if (outbound) {
                 acknowledged.set(true)
             }
         }
--- a/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
@@ -20,9 +20,10 @@ import java.util.concurrent.atomic.AtomicInteger
 import kotlin.math.max
 import kotlin.properties.Delegates
 
-const val ACK_BACKLOG_ALLOWANCE = 256
-const val INITIAL_WINDOW_SIZE = 256 * 1024
 const val DEFAULT_MAX_BUFFERED_CONNECTION_WRITES = 10 * 1024 * 1024 // 10 MiB
+const val DEFAULT_ACK_BACKLOG_LIMIT = 256
+
+const val INITIAL_WINDOW_SIZE = 256 * 1024
 
 open class YamuxHandler(
     override val multistreamProtocol: MultistreamProtocol,
@@ -31,6 +32,7 @@ open class YamuxHandler(
     inboundStreamHandler: StreamHandler<*>,
     private val connectionInitiator: Boolean,
     private val maxBufferedConnectionWrites: Int,
+    private val ackBacklogLimit: Int,
     private val initialWindowSize: Int = INITIAL_WINDOW_SIZE
 ) : MuxHandler(ready, inboundStreamHandler) {
 
@@ -217,9 +219,7 @@ open class YamuxHandler(
     }
 
     override fun onLocalOpen(child: MuxChannel<ByteBuf>) {
-        if (totalUnacknowledgedStreams() < ACK_BACKLOG_ALLOWANCE) {
-            createYamuxStreamHandler(child.id).onLocalOpen()
-        }
+        createYamuxStreamHandler(child.id).onLocalOpen()
     }
 
     private fun onRemoteYamuxOpen(id: MuxId) {
@@ -243,6 +243,13 @@ open class YamuxHandler(
 
     override fun onChildClosed(child: MuxChannel<ByteBuf>) {
         streamHandlers.remove(child.id)?.dispose()
+    }
+
+    override fun checkCanOpenNewStream() {
+        val totalUnacknowledgedStreams = streamHandlers.values.count { !it.acknowledged.get() }
+        if (totalUnacknowledgedStreams >= ackBacklogLimit) {
+            throw AckBacklogLimitExceededMuxerException("The ACK backlog limit of $ackBacklogLimit streams has been reached. Will not open new stream.")
+        }
     }
 
     private fun handlePing(msg: YamuxFrame) {
@@ -270,10 +277,6 @@ open class YamuxHandler(
 
     private fun calculateTotalBufferedWrites(): Int {
         return streamHandlers.values.sumOf { it.sendBuffer.readableBytes() }
-    }
-
-    private fun totalUnacknowledgedStreams(): Int {
-        return streamHandlers.values.count { !it.acknowledged.get() }
     }
 
     override fun generateNextId() =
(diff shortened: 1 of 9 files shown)


mxinden:
//CC @thomaseizinger who pushed a lot for this to land in rust-libp2p and specs.

Nashatyrev:
Just to make sure I'm getting the Yamux spec correctly: 

There could be two kinds of unacknowledged streams: 
- outbound: local peer opens a stream, sends `SYN`, but hasn't received `ACK` yet
- inbound: in our current implementation we are sending back `ACK` immediately upon receiving a `SYN` from a remote party. In that case there is no way to have unacknowledged inbound streams. However the spec states the following: 
`MAY delay acknowledging new streams until the application has received or is about to send the first DATA frame`.
If this statement is implemented at some point then a remotely opened stream may also be unacknowledged. I.e. until the first inbound data frame is received or a local app starts sending a data. 

So the questions: 
1. should we account both types as a sum and treat them equally?
2. In the case of exceeding unacknowledged streams threshold what should we do when ... opening a new stream locally? Should we throw `MuxerWriteException`? (i think yes) Should we also abort the connection? (probably no) 
3. In the case of exceeding unacknowledged streams threshold what should we do when ... receiving `SYN` from remote? Abort the stream (by sending back `RST`)? Abort the connection?

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -10,19 +10,17 @@ import io.libp2p.etc.types.writeOnce
...
 import io.netty.buffer.ByteBuf
 import io.netty.channel.ChannelHandlerContext
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicInteger
 import kotlin.math.max
 import kotlin.properties.Delegates
 
+const val ACK_BACKLOG_ALLOWANCE = 256
```
Nashatyrev, line 23: I would parametrize it for the `YamuxHandler` making this constant default value

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -214,7 +217,9 @@ open class YamuxHandler(
     }
 
     override fun onLocalOpen(child: MuxChannel<ByteBuf>) {
-        createYamuxStreamHandler(child.id).onLocalOpen()
+        if (totalUnacknowledgedStreams() < ACK_BACKLOG_ALLOWANCE) {
+            createYamuxStreamHandler(child.id).onLocalOpen()
+        }
```
Nashatyrev, line 222: Should we throw a specific `WriteMuxerException` here?

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -214,7 +217,9 @@ open class YamuxHandler(
...
     }
 
     private fun onRemoteYamuxOpen(id: MuxId) {
```
Nashatyrev, line 225: Do we need to check `ACK_BACKLOG_ALLOWANCE` here as well?

Nashatyrev (commented):
Some questions remain

mxinden:
> should we account both types as a sum and treat them equally?

They should be accounted separately. A should not ACK inbound streams from B, in case A's application layer has not yet read or written to a stream. B should not open a new stream to A in case 256 outbound streams to A have not yet been ACKed. Same vice versa for streams from A to B.

mxinden:
> In the case of exceeding unacknowledged streams threshold what should we do when ... opening a new stream locally? Should we throw `MuxerWriteException`? (i think yes) Should we also abort the connection? (probably no)

Do you have any way to backpressure the application? In Rust and I am assuming in Go as well, we suspend the green thread.

mxinden:
> In the case of exceeding unacknowledged streams threshold what should we do when ... receiving `SYN` from remote? Abort the stream (by sending back `RST`)? Abort the connection?

In rust-yamux today it is best effort only, enforced at the outbound end for now. 

https://github.com/libp2p/rust-yamux/blob/93b70626f7f3d83f97951d9cd5c2951500088f0d/yamux/src/connection.rs#L435-L439

In case you want to enforce it at the receiving side, I would consider resetting the stream exceeding the limit.

Nashatyrev:
> Do you have any way to backpressure the application? 

Not with the present API 
> In Rust and I am assuming in Go as well, we suspend the green thread.

Yeah, that's a good solution. 
> In case you want to enforce it at the receiving side, I would consider resetting the stream exceeding the limit.

If both peers has the same `ACK_BACKLOG_ALLOWANCE` then this situation should be treated as the protocol violation. Am I getting it right? 
In that case even connection abort would be appropriate

mxinden:
> > In case you want to enforce it at the receiving side, I would consider resetting the stream exceeding the limit.
> 
> If both peers has the same `ACK_BACKLOG_ALLOWANCE` then this situation should be treated as the protocol violation. Am I getting it right?

Correct. Though I am not aware of a single libp2p based network out there today that supports the yamux ack backlog feature across all nodes. Thus treating it as best effort everywhere.

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -240,8 +245,11 @@ open class YamuxHandler(
         streamHandlers.remove(child.id)?.dispose()
     }
 
-    private fun calculateTotalBufferedWrites(): Int {
-        return streamHandlers.values.sumOf { it.sendBuffer.readableBytes() }
+    override fun checkCanOpenNewStream() {
```
Nashatyrev, line 248: Why not just throw from `onLocalOpen()` ?

StefanBratanov: So essentially in `newStream` method in `AbstractMuxHandler`, if `onLocalOpen` fails then `outboundInitializer` is never called and `createStream` in `MuxHandler` controller future is never complete. And I am thinking also in order to avoid calling `createChild` with he new id.

Nashatyrev: With the fix above throwing from `onLocalOpen()` worked for me

StefanBratanov: Nice. I like it.

Nashatyrev:
> It also looks like we need to treat initiator/responder acknowledgments separately as per comment from @mxinden 

@StefanBratanov are you planning to address this here yet or later? 
Else the PR looks good to me.

StefanBratanov:
> > It also looks like we need to treat initiator/responder acknowledgments separately as per comment from @mxinden
> 
> @StefanBratanov are you planning to address this here yet or later? Else the PR looks good to me.

Working on it now. Think it makes sense to be in this PR as well.

StefanBratanov:
> > > It also looks like we need to treat initiator/responder acknowledgments separately as per comment from @mxinden
> > 
> > 
> > @StefanBratanov are you planning to address this here yet or later? Else the PR looks good to me.
> 
> Working on it now. Think it makes sense to be in this PR as well.

Done.

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -87,11 +91,23 @@ open class YamuxHandler(
                     writeAndFlushFrame(YamuxFrame(msg.id, YamuxType.WINDOW_UPDATE, YamuxFlag.ACK.asSet, 0))
                 }
 
+                YamuxFlag.ACK in msg.flags -> {
+                    if (outbound) {
+                        acknowledged.set(true)
+                    }
```
Nashatyrev, line 97: nit: May be extract to function `acknowledgeOutboundStreamIfNeeded()` for uniformity?

libp2p/src/main/kotlin/io/libp2p/mux/yamux/YamuxHandler.kt
```diff
@@ -87,11 +91,23 @@ open class YamuxHandler(
...
+                }
+
                 YamuxFlag.FIN in msg.flags -> onRemoteDisconnect(msg.id)
                 YamuxFlag.RST in msg.flags -> onRemoteClose(msg.id)
             }
         }
 
+        private fun acknowledgeInboundStreamIfNeeded() {
+            if (!outbound) {
+                acknowledged.compareAndSet(false, true)
```
Nashatyrev, line 107: nit: just `set()` seems to yield the same result?
BTW: I believe we basically don't need any thread safe code here. All the methods should be invoked on the same Netty event thread. Anyways the overall `YamuxHandler` code is not thread safe globally.

Nashatyrev (approved):
LGTM, just some nits 👍

PR #343, opened 2023-11-13
Don't try and dial DNSADDR addresses

ianopolous:
These should mainly be used for bootstrap with a DNS TXT record lookup to expand to list of node multiaddrs

This removes a lot of log spam from failed dial attempts in kademlia


ianopolous:
Any thoughts on this @Nashatyrev ?

Nashatyrev (approved):
Looks good to me! 

It feels like there should be more generic handling, but for now I'm fine with this additional 'if' 

P.S. sorry for ignoring your PR, I seem to be subscribed for _all_ repository notifications but for some reasons get only those where I mentioned explicitly. So feel free just to CC me on any new PRs

Nashatyrev:
This PR is the follow up for #334

Issue #344, opened 2023-11-24
circuit-relay-v2


ianopolous:
Hi @Nashatyrev et al,

I'm working on an implementation of circuit-relay-v2 in another repo and would like to upstream it, if you are interested, when it is working. Is this something you'd like upstreamed?

I've got the relay working, and relaying connections. I'm just working on adding the security and muxer upgrade to relayed streams. I've implemented a new RelayTransport and made sure TcpTransport won't try and handle circuit addresses. 

What are you thoughts on the direction of this: https://github.com/Peergos/nabu/compare/feat/circuit-relay-v2?expand=1

The important bit is HostBuilder::upgradeConnection for handling incoming relayed connections. Once this is done I'll work on dcutr.

Nashatyrev:
Hey @ianopolous 

> What are you thoughts on the direction of this: https://github.com/Peergos/nabu/compare/feat/circuit-relay-v2?expand=1

That doesn't look like a lot of changes to the core classes there. I believe it's worth to upstream. 
@StefanBratanov what do you think?

ianopolous:
@Nashatyrev It's looking like I might need more upstream changes than I was hoping. Any suggestions for how to get RelayTransport::upgradeStream working when we have a Stream not a Channel? It seems we fundamentally need a channel to set the remote peerid on for the security negotiation?

StefanBratanov:
> Hey @ianopolous
> 
> > What are you thoughts on the direction of this: https://github.com/Peergos/nabu/compare/feat/circuit-relay-v2?expand=1
> 
> That doesn't look like a lot of changes to the core classes there. I believe it's worth to upstream. @StefanBratanov what do you think?

I think it's good addition especially since there is already a project which would require it and most of the code is already there.

Nashatyrev:
@ianopolous, sorry for delay again, I'm OOO currently 
I'll try to look into this in a week or so

ianopolous:
My hope is to test with a local relay once it's working. But wanted to rule out issues with the relay first. The normal way to find a relay is to try a bunch of nodes you've discovered, as it is on by default normally. We do this downstream in Nabu using our kademlia impl.

ianopolous:
I've changed the test relay to another random node.

ianopolous:
I've tried calling channel.pipeline().fireChannelActive() in a bunch of places including in upgradeStream after calling upgrader.establishSecureChannel(). None of it seems to have any effect.

[...]

PR #345, opened 2023-12-31
Implement circuit relay v2

ianopolous:
Specs:
https://docs.libp2p.io/concepts/nat/circuit-relay/
https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md

Go source: https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go

Fixes https://github.com/libp2p/jvm-libp2p/issues/344


libp2p/src/main/java/io/libp2p/protocol/circuit/crypto/pb/Crypto.java
```diff
@@ -0,0 +1,2441 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
Nashatyrev, line 1: Could you please add `*.proto` files and remove generated classes from the PR?

libp2p/src/main/kotlin/io/libp2p/core/multiformats/Multiaddr.kt
```diff
@@ -61,7 +61,11 @@ data class Multiaddr(val components: List<MultiaddrComponent>) {
      * @throws IllegalArgumentException if existing component value doesn't match [value]
      */
     private fun withComponentImpl(protocol: Protocol, value: ByteArray?): Multiaddr {
-        val existingComponent = getFirstComponent(protocol)
+        val existingComponent = if (has(Protocol.P2PCIRCUIT)) {
+            split { it == Protocol.P2PCIRCUIT }.get(1).getFirstComponent(protocol)
```
Nashatyrev, line 65: I think those existing `withComponent` methods hit their limitations with introducing `relay` protocol. I believe we would need to introduce a builder for `Multiaddr` which would be able to construct/modify components in a more explicit way.   
This could be addressed later in a separate PR

libp2p/src/main/java/io/libp2p/protocol/circuit/RelayTransport.java
```diff
@@ -0,0 +1,334 @@
...
+                ensureEnoughCurrentRelays();
+                try {
+                  Thread.sleep(2 * 60_000);
+                } catch (InterruptedException e) {
+                }
+              }
+            },
+            "RelayManager")
+        .start();
+  }
```
Nashatyrev, line 260: I would use `ScheduledExecutorService` here which is explicitly passed via `RelayTransport` constructor

Nashatyrev:
Basically looks good to me! 
I didn't dive deeply into the circuit implementation code though 
I would be happy if you could add some (at least sanity) unit tests for protocol implementations and 
if possible add the `RelayTransport` test as a subclass of `TransportTests`

Could you also please add a 🍋 for the `Circuit Relay` component row on the main README.md 😄

ianopolous:
It seems like the protobuf changes have broken a peerid test. It's not clear to me why yet.

ianopolous:
I had a look at making RelayTransportTest a subclass of TransportTests, but it seems quite awkward, because it needs the Host in the transport via setHost and the host builder modifier. It would be nice to find a better solution to this that didn't need to use HostBuilder().builderModifier()

Nashatyrev (approved):
LGTM

@StefanBratanov may be you could give another review?

StefanBratanov:
> LGTM
> 
> @StefanBratanov may be you could give another review?

I am having a look.

StefanBratanov (approved):
LGTM to me as well

PR #348, opened 2024-01-23
adds example gossip pubsub with discovery

philrhc:
I've had a bit of trouble adding gossip pubsub with discovery into an application, so i wanted to establish an example. Here is a very very simple app to connect publisher and subscriber, printing the parsed counter


philrhc:
nvm

PR #354, opened 2024-03-05
Add propagator in `MessageApi` when subscribed to a pubsub topic

StefanBratanov:
Add `val propagator` to `MessageApi` which can be used by downstream subscribers to know who has propagated the gossip message


StefanBratanov:
Closing this because alternative solutions were discussed

PR #364, opened 2024-05-15
Fix remotePubKey in Noise secure Session

Nashatyrev:
Noise negotiator was putting the wrong pubKey into `SecureChannel.Session`: local instead of remote

- Fix the issue
- Add a unit test


StefanBratanov (approved):
LGTM

Issue #361, opened 2024-04-23
Possible Bug on GossipSub implementation that makes sharing `IHAVE` control messages with empty topics


cortze:
### Summary

After streaming libp2p traces from our ([ProbeLab](https://github.com/probe-lab) team) custom libp2p-trace generator tool [`Hermes` ](https://github.com/probe-lab/hermes), we’ve spotted that there are some peers in the Ethereum network sharing `IHAVE` messages, including `msg_ids`, on empty topics.

![image](https://github.com/libp2p/jvm-libp2p/assets/45786396/887d5320-ce13-4d2a-8fbd-269cfc59d08d)

After inspecting this behaviour closely and double-checking that this isn’t a problem or a bug on our end, we’ve been able to verify that:

- our node is not subscribed to an empty topic
- our node isn’t sending any control messages on an empty topic
- the shared `message_ids` on those `IHAVE` messages with empty topics do belong to other complete Ethereum topics

After checking any existing correlation between the remote peerIDs sending those messages and any particular client or libp2p implementation, we could count a total of 55 unique peers that sent those empty-topic messages over 3.5 hours, where 49/55 of those peers were successfully identified as `teku/teku` nodes, and the 6 missing peers were unable to be identified.

There hasn’t been any further investigation on which could be the root problem that generates those buggy `IHAVE` control messages, but it seems that this could be related to an implementation bug, where the topics aren’t mapped correctly to `msg_ids` or aren’t correctly added to the control RPC.

Let us know if there is anything else we can help you with :)

### Expected behavior

Incoming control GossipSub messages should:
- first, include only topics that we are subscribed to ↔ we weren’t subscribed to an empty topic
- Shouldn’t advertise `msg_ids` from other topics on the wrong topics

### Actual behavior

These are trace examples read by the `Hermes` tool:

```
# Traced message
{
	'Type': 'RECV_RPC', 
	'PeerID': '16Uiu2HAm5BX98oN6Z8Rb...StKkyrNdNsdKZyw4hXutE', 
	'Timestamp': '2024-04-11T19:53:25.725929846+02:00', 
	'Data': {
		'PeerID': '16Uiu2HAmVfzYfKQvjqY...2oouvW1XVAF7eKttz9',
		'Control': {
			'IHave': [{
				'TopicID': '', 
[...]

PR #365, opened 2024-05-15
Set topicID on outbound IHAVE and ignore inbound IHAVE for unknown topic

StefanBratanov:
fixes https://github.com/libp2p/jvm-libp2p/issues/361


Nashatyrev:
I think we would need to group `IHAVE`s in the resulting message produced by `GossipRpcPartsQueue` else we might get significant message size overhead

Nashatyrev:
Please check here: https://github.com/StefanBratanov/jvm-libp2p/pull/1
Fixing the case has also addressed the grouping question

StefanBratanov:
> Please check here: [StefanBratanov#1](https://github.com/StefanBratanov/jvm-libp2p/pull/1) Fixing the case has also addressed the grouping question

Thanks. I agree with the change. I have missed that part. Thanks.

Nashatyrev:
Added another minor improvement

Nashatyrev (approved):
LGTM

PR #366, opened 2024-05-20
Fix mdns when listening with ipv6 wildcard (which includes ipv4)

ianopolous:
The listen port method was throwing NPE


Nashatyrev (approved):
LGTM
Thanks for the fix!

PR #368, opened 2024-05-22
More mdns fixes

ianopolous:
This fixes mdns discovery when using wildcard listen addresses.

It also adds a method to enable adding PeerListener from java (CopyOnWriteList isn't helpful)

Also allow ipv6 only peers

@Nashatyrev


Nashatyrev:
@StefanBratanov maybe you have a look?

StefanBratanov (approved):
LGTM

PR #375, opened 2024-09-25
Fix Gossip simulator issue 

Nashatyrev:
The message delayer was delaying a subsequent message till the previous message is received. 
That included latency which is wrong


StefanBratanov (approved):
LGTM

PR #385, opened 2024-10-16
Don't throw NoPeersForOutboundMessageException if peers DONTWANT message

StefanBratanov:
If all peers for a certain topic have sent IDONTWANT for a messageId, then when we publish it outbound, we don't want NoPeersForOutboundMessageException to be thrown

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -419,9 +419,10 @@ open class GossipRouter(
         val list = peers
             .map {
                 if (peerDoesNotWantMessage(it, msg.messageId)) {
-                    return CompletableFuture.completedFuture(null)
+                    CompletableFuture.completedFuture(Unit)
+                } else {
+                    submitPublishMessage(it, msg)
                 }
-                return submitPublishMessage(it, msg)
             }
 
         mCache += msg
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -416,15 +416,18 @@ open class GossipRouter(
...
-            anyComplete(list)
+        return if (peers.isNotEmpty()) {
+            val publishedMessages = peers
+                .filterNot { peerDoesNotWantMessage(it, msg.messageId) }
+                .map { submitPublishMessage(it, msg) }
+            if (publishedMessages.isEmpty()) {
+                return CompletableFuture.completedFuture(Unit)
+            }
+            flushAllPending()
+            anyComplete(publishedMessages)
```
Nashatyrev, line 430: nit: Could you please just rearrange this code a bit: 
```suggestion
            if (publishedMessages.isEmpty()) {
                CompletableFuture.completedFuture(Unit)
            } else {
                flushAllPending()
                anyComplete(publishedMessages)
            }
```

Nashatyrev (approved):
Yeah , great catch 👍 
LGTM

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
StefanBratanov: Done

PR #386, opened 2024-10-18
Send IDONTWANT prior to publish

### LINKED ISSUE #383: Gossip: send `IDONTWANT` selectively prior to publish
### Description

There are cases when the same message could be published by more than one peer. 
In such cases it would be nice to broadcast `IDONTWANT` right prior to publishing. 

### Motivation

E.g. Ethereum DAS data column may be reseeded to a corresponding topic in the following cases: 
- A column was restored via KZG from 50% of other data columns 
- The data was pulled ahead from the Execution Layer transaction pool 

### Requirements

Gossip publish API needs to be modified 

### Open questions

_No response_

### Are you planning to do it yourself in a pull request ?

Maybe

StefanBratanov:
Send IDONTWANT on each publish. Maybe we need to add some filtering to make it more selective, but not sure?

I removed filtering out excluding the `receivedFrom` peer when we are sending IDONTWANT on inbound messages. Think it simplifies the flow and could help in some cases, where `receivedFrom` may send the same message twice.

fixes #383

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -166,7 +166,7 @@ open class GossipRouter(
     }
 
     override fun notifyUnseenMessage(peer: PeerHandler, msg: PubsubMessage) {
-        iDontWant(msg, peer)
+        iDontWant(msg)
         eventBroadcaster.notifyUnseenMessage(peer.peerId, msg)
         notifyAnyMessage(peer, msg)
     }
@@ -420,6 +420,7 @@ open class GossipRouter(
         mCache += msg
 
         return if (peers.isNotEmpty()) {
+            iDontWant(msg)
             val publishedMessages = peers
                 .filterNot { peerDoesNotWantMessage(it, msg.messageId) }
                 .map { submitPublishMessage(it, msg) }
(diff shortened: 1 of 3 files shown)


libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -610,15 +611,14 @@ open class GossipRouter(
         enqueueIwant(peer, messageIds)
     }
 
-    private fun iDontWant(msg: PubsubMessage, receivedFrom: PeerHandler) {
+    private fun iDontWant(msg: PubsubMessage) {
```
Nashatyrev, line 614: TBH I would leave the excluded peer. We are saving one IDONTWANT message for free. 
```suggestion
    private fun iDontWant(msg: PubsubMessage?) {
```

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -610,15 +611,14 @@ open class GossipRouter(
...
         if (!this.protocol.supportsIDontWant()) return
         if (msg.protobufMessage.data.size() < params.iDontWantMinMessageSizeThreshold) return
         // we need to send IDONTWANT messages to mesh peers immediately in order for them to have an effect
         msg.topics
             .mapNotNull { mesh[it] }
             .flatten()
             .distinct()
-            .minus(receivedFrom)
```
Nashatyrev, line 621:
```kotlin
            .minus(setOfNotNull(receivedFrom))
```

Nashatyrev (approved):
LGTM, I'd just leave the peer exclusion

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
StefanBratanov: Oh, I was looking for a nice way to add it back. This one seems good. Implemented.

PR #387, opened 2024-10-18
Gossip: more reliable publishing  

Nashatyrev:
The quote from the [Gossip spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#message-processing):
> - If the router is subscribed to the topic, it will send the message to all peers in mesh[topic].
> - If the router is not subscribed to the topic ...

According to spec we should forward the message we are publishing to the mesh peers only. 

However the mesh could be not that stable (for example a peer may be evicted from mesh due to slight gossip downscorring (any score `<0`)). Thus it could lead to the situation, when the mesh is empty but there are still peers subscribed to the topic outside of the mesh. According to the spec we are not able to publish. 

The situation could not be that dramatic due to `IHAVE/IWANT` mechanism which should finally make the message published, but that could induce significant publish latency. 

To overcome this it does make sense to take extra peers outside of the mesh when the mesh size is `< D`. 

This is also the subject for a PR to the Gossip spec

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -408,11 +408,20 @@ open class GossipRouter(
                     .plus(getDirectPeers())
             } else {
                 msg.topics
-                    .mapNotNull { topic ->
-                        mesh[topic] ?: fanout[topic] ?: getTopicPeers(topic).shuffled(random).take(params.D)
-                            .also {
-                                if (it.isNotEmpty()) fanout[topic] = it.toMutableSet()
-                            }
+                    .map { topic ->
+                        val topicMeshPeers = mesh[topic]
+                        if (topicMeshPeers != null) {
+                            // we are subscribed to the topic
+                            val addFromNonMeshCount = max(0, params.D - topicMeshPeers.size)
+                            val nonMeshTopicPeers = (getTopicPeers(topic) - topicMeshPeers)
+                            topicMeshPeers + nonMeshTopicPeers.shuffled(random).take(addFromNonMeshCount)
+                        } else {
+                            // we are not subscribed to the topic
+                            fanout[topic] ?: getTopicPeers(topic).shuffled(random).take(params.D)
+                                .also {
+                                    if (it.isNotEmpty()) fanout[topic] = it.toMutableSet()
+                                }
+                        }
                     }
                     .flatten()
             }
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -413,7 +413,7 @@ open class GossipRouter(
                         if (topicMeshPeers != null) {
                             // we are subscribed to the topic
                             val addFromNonMeshCount = max(0, params.D - topicMeshPeers.size)
-                            val nonMeshTopicPeers = (getTopicPeers(topic) - topicMeshPeers)
+                            val nonMeshTopicPeers = getTopicPeers(topic) - topicMeshPeers
                             topicMeshPeers + nonMeshTopicPeers.shuffled(random).take(addFromNonMeshCount)
                         } else {
                             // we are not subscribed to the topic
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -404,8 +404,10 @@ open class GossipRouter(
     override fun broadcastOutbound(msg: PubsubMessage): CompletableFuture<Unit> {
         msg.topics.forEach { lastPublished[it] = currentTimeSupplier() }
 
+        val floodPublish = msg.size <= params.floodPublishMaxMessageSizeThreshold
+
         val peers =
-            if (params.floodPublish) {
+            if (floodPublish) {
                 msg.topics
                     .flatMap { getTopicPeers(it) }
                     .filter { score.score(it.peerId) >= scoreParams.publishThreshold }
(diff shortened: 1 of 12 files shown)


tbenr:
Can't find the definition of "direct peer" in the spec, but per my mental model two direct peers mutually configured as "direct peer" always exchange messages among each other.
If we always send messages direct peers when dealing with `broadcastInbound`, I don't see why we shouldn't do the same when publishing locally produced messages via `broadcastOutbound`.

Then I also wonder why direct peers are blindly considered for outbound even if they don't subscribe to the specific topic (but this is another question :-), and maybe it's fine)

tbenr (approved):
LGTM

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -408,11 +408,20 @@ open class GossipRouter(
...
-                        mesh[topic] ?: fanout[topic] ?: getTopicPeers(topic).shuffled(random).take(params.D)
-                            .also {
-                                if (it.isNotEmpty()) fanout[topic] = it.toMutableSet()
-                            }
+                    .map { topic ->
+                        val topicMeshPeers = mesh[topic]
+                        if (topicMeshPeers != null) {
+                            // we are subscribed to the topic
+                            val addFromNonMeshCount = max(0, params.D - topicMeshPeers.size)
+                            val nonMeshTopicPeers = (getTopicPeers(topic) - topicMeshPeers)
```
StefanBratanov, line 416: Nit: no need of the brackets

StefanBratanov (approved):
LGTM to me as well

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
StefanBratanov, line 416: Do we care about the score of the topic peers in this case?

tbenr: > Do we care about the score of the topic peers in this case?

I think the scoring is already "captured" by the mesh. Outside that, seems reasonable to me to pick random peers. 

On Friday I had a similar comment suggesting "why don't we select by sorting them by score" and Anton had a good reply to it.

Nashatyrev: Yeah, good question. Here is the [quotation](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#score-thresholds) from the spec: 
> `PublishThreshold`: when a peer's score drops below this threshold, self-published messages are not propagated towards this peer when (flood) publishing. This threshold should be negative, and less than or equal to the gossip threshold.

I believe this mean that both 'flood' and 'regular' publishing should consider this threshold. 
It's probably counter eclipse attack measure. 
So I believe we still need to truncate those with the score `< PublishThreshold` _but_ still consider them when there are few peers to publish

Does it make sense?

Nashatyrev:
> Can't find the definition of "direct peer" in the spec, but per my mental model two direct peers mutually configured as "direct peer" always exchange messages among each other. If we always send messages direct peers when dealing with `broadcastInbound`, I don't see why we shouldn't do the same when publishing locally produced messages via `broadcastOutbound`.
> 
> Then I also wonder why direct peers are blindly considered for outbound even if they don't subscribe to the specific topic (but this is another question :-), and maybe it's fine)

@tbenr you are right. There is a mess with direct peers. Just created a separate issue for this: https://github.com/libp2p/jvm-libp2p/issues/389

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
tbenr: Yes that seems like a convincing interpretation to me.

Nashatyrev:
Addressed the peer score question when publishing. 
@StefanBratanov @tbenr could you please take another look please?

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -408,11 +412,28 @@ open class GossipRouter(
...
+                            val (nonMeshTopicPeersAbovePublishThreshold, nonMeshTopicPeersBelowPublishThreshold) =
+                                nonMeshTopicPeers.partition { score.score(it.peerId) >= scoreParams.publishThreshold }
+                            // this deviates from the original spec but we want at least D peers for publishing
+                            // prioritizing mesh peers, then non-mesh peers with acceptable score,
+                            // and then underscored non-mesh peers as a last resort
+                            (
+                                topicMeshPeers +
+                                    nonMeshTopicPeersAbovePublishThreshold.shuffled(random) +
+                                    nonMeshTopicPeersBelowPublishThreshold.shuffled(random)
+                                ).take(params.D)
```
tbenr, line 429: the logic makes sense to me, but I don't like that we and up always doing the partitioning and shuffling even if the topic fulfill the `D` param (which is supposed to be the most common case). Maybe we can go for it only when topicPeers is < `D`?

Nashatyrev: Yep, just made it alongside with fixing a regression

libp2p/src/test/kotlin/io/libp2p/pubsub/gossip/GossipV1_1Tests.kt
```diff
@@ -1128,34 +1129,208 @@ class GossipV1_1Tests : GossipTestsBase() {
...
+    @Test
+    fun `should publish to all mesh peers when mesh exceeds D`() {
+        val gossipParams = GossipParams(D = 6, DHigh = 10)
+        val test = ManyRoutersTest(params = gossipParams, mockRouterCount = gossipParams.DHigh)
+        val topic = "topic1"
+        test.connectAll()
+
+        test.mockRouters.forEach {
+            it.subscribe(topic)
+//            it.sendToSingle(createGraftMessage(topic))
```
tbenr, line 1181: we can delete this :-)

tbenr (approved):
LGTM

PR #391, opened 2024-10-21
[GOSSIP][BREAKING] Replace `floodPublish` param with `floodPublishMaxMessageSizeThreshold` 

### LINKED ISSUE #381: Consider converting floodPublish Gossip parameter to message size based
### Description

Introduce `flood_publish_max_size: Long` option instead of `flood_publish: Boolean`
'flood publish' feature would be used just for the messages of size above the specified threshold 
former `flood_publish = false` would be equivalent to `flood_publish_max_size = 0`
former `flood_publish = true` would be equivalent to `flood_publish_max_size = Long.MAX_VALUE`

### Motivation

`floodPublish` ([spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing)) is a good feature to mitigate eclipse attacks against a publisher from one hand. 
From the other hand it may significantly delay message publishing. Especially messages of larger size with limited outbound bandwidth. 
For example let's consider publishing a 1Mb message with bandwidth of 100Mbit/sec. Publishing it without `floodPublish` feature would take roughly **1sec** with mesh size of 10 peers. Publishing it with `floodPublish` feature with 100 peers on the topic would take roughly **10sec**. 
So instead of completely disabling `floodPublish` for limited bandwidth setups it makes sense to apply the feature depending on the message size. So the messages of the size above the threshold would be sent without utilizing `floodPublish`. 
The former 

### Current Implementation

The `floodPublish` parameter is currently `boolean` and is capable of just totally switching on/off flood publish feature 

### Are you planning to do it yourself in a pull request ?

Maybe

tbenr:
fixes #381

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
@@ -22,6 +22,10 @@ fun defaultDLazy(D: Int) = D
 fun defaultDScore(D: Int) = D * 2 / 3
 fun defaultDOut(D: Int, DLow: Int) = min(D / 2, max(DLow - 1, 0))
 
+// floodPublishMaxMessageSizeThreshold shortcuts
+const val NEVER_FLOOD_PUBLISH = 0
+const val ALWAYS_FLOOD_PUBLISH = Int.MAX_VALUE
+
 /**
  * Parameters of Gossip 1.1 router
  */
@@ -116,9 +120,12 @@ data class GossipParams(
      * published using flood publishing mode.
      * When a message size is <= [floodPublishMaxMessageSizeThreshold], published messages are forwarded
      * to all peers with score >= to [GossipScoreParams.publishThreshold]
-     * The default is 0 KiB (never flood publish).
+     *
+     * [NEVER_FLOOD_PUBLISH] and [ALWAYS_FLOOD_PUBLISH] can be used as shortcuts.
+     *
+     * The default is [NEVER_FLOOD_PUBLISH] (0 KiB).
      */
-    val floodPublishMaxMessageSizeThreshold: Int = 0,
+    val floodPublishMaxMessageSizeThreshold: Int = NEVER_FLOOD_PUBLISH,
 
     /**
      * [gossipFactor] affects how many peers we will emit gossip to at each heartbeat.
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
@@ -112,11 +112,13 @@ data class GossipParams(
     val seenTTL: Duration = 2.minutes,
 
     /**
-     * [floodPublish] is a gossipsub router option that enables flood publishing.
-     * When this is enabled, published messages are forwarded to all peers with score >=
-     * to publishThreshold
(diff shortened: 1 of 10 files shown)


libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
```diff
@@ -400,8 +400,10 @@ open class GossipRouter(
     override fun broadcastOutbound(msg: PubsubMessage): CompletableFuture<Unit> {
         msg.topics.forEach { lastPublished[it] = currentTimeSupplier() }
 
+        val floodPublish = msg.protobufMessage.data.size() <= params.floodPublishMaxMessageSizeThreshold
```
Nashatyrev, line 403: We already refer to 'message size' in a couple of places. I would add this notion to `PubsubMessage` interface: 
```patch
Index: libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubRouter.kt
IDEA additional info:
<+>UTF-8
diff --git a/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubRouter.kt b/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubRouter.kt
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubRouter.kt	(revision 1defc232bebd9bb3eaafd05be1e7fa14f436f328)
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubRouter.kt	(date 1729512008659)
@@ -22,6 +22,9 @@
     val topics: List<Topic>
         get() = protobufMessage.topicIDsList
 
+    val size: Long
+        get() = protobufMessage.data.size().toLong()
+
     fun messageSha256() = sha256(protobufMessage.toByteArray())
 
     override fun equals(other: Any?): Boolean
```

Nashatyrev (approved):
LGTM,
just several nits

Nashatyrev:
Oh, would be great to have a unit test(s) for this parameter

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
```diff
@@ -112,11 +112,13 @@ data class GossipParams(
...
-     * When this is enabled, published messages are forwarded to all peers with score >=
-     * to publishThreshold
+     * [floodPublishMaxMessageSizeThreshold] controls the maximum size (in bytes) a message will be
+     * published using flood publishing mode.
+     * When a message size is <= [floodPublishMaxMessageSizeThreshold], published messages are forwarded
+     * to all peers with score >= to [GossipScoreParams.publishThreshold]
+     * The default is 0 KiB (never flood publish).
      */
-    val floodPublish: Boolean = false,
+    val floodPublishMaxMessageSizeThreshold: Int = 0,
```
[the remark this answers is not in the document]
tbenr: yes that makes sense

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
tbenr: right, yes.. seems like you like the idea to expose `Long` even if under the hood we can't go over 2gb (protobuf limit?)

Nashatyrev: Oh, it may be an overkill. Let's stick with just Int then

PR #392, opened 2024-10-21
Logging and other small warnings removal

tbenr:
I don't know if everything is correct here, I'm not a kotlin expert

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -23,7 +23,7 @@ const val DEFAULT_MAX_PUBSUB_MESSAGE_SIZE = 1 shl 20
 
 typealias PubsubMessageHandler = (PubsubMessage) -> CompletableFuture<ValidationResult>
 
-class DefaultPubsubMessage(override val protobufMessage: Rpc.Message) : AbstractPubsubMessage() {
+open class DefaultPubsubMessage(override val protobufMessage: Rpc.Message) : AbstractPubsubMessage() {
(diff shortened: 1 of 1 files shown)


tbenr:
@Nashatyrev @StefanBratanov

libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
```diff
@@ -23,7 +23,7 @@ const val DEFAULT_MAX_PUBSUB_MESSAGE_SIZE = 1 shl 20
 
 typealias PubsubMessageHandler = (PubsubMessage) -> CompletableFuture<ValidationResult>
 
-open class DefaultPubsubMessage(override val protobufMessage: Rpc.Message) : AbstractPubsubMessage() {
+class DefaultPubsubMessage(override val protobufMessage: Rpc.Message) : AbstractPubsubMessage() {
```
Nashatyrev, line 26: Until this project has strict rules regarding members visibility I would suggest to leave them as is. I was using some of them for tweaking router behavior in side projects.

libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
```diff
@@ -105,7 +104,7 @@ abstract class AbstractRouter(
     }
 
     override fun addNewStream(stream: Stream) = addNewStreamWithHandler(stream, null)
-    protected fun addNewStreamWithHandler(stream: Stream, handler: ChannelHandler?) {
+    private fun addNewStreamWithHandler(stream: Stream, handler: ChannelHandler?) {
```
Nashatyrev, line 107: Same here

Nashatyrev (approved):
LGTM, but I would ask to keep the stuff open for hacking

PR #407, opened 2025-04-02
Quic secure transport implementation

ianopolous:
Also support webrtc-direct multiaddrs

Quic is working fully java-java and java - kubo (go-libp2p) and java - rust-libp2p.

Trying to use Ed25519 certs in the TLS part of quic fails with a cert error. I suspect this is a Netty bug, but it shouldn't block merging as everyone seems to recommend just using ECDSA. 

@Nashatyrev 

I'm not sure what the Android issue is in CI, but I don't think it's related to these changes.


Nashatyrev:
For me looks good from the initial glance 👍 
Pinging Teku team: @StefanBratanov @tbenr

libp2p/build.gradle.kts
```diff
@@ -14,6 +14,8 @@ dependencies {
     api("io.netty:netty-transport")
     implementation("io.netty:netty-handler")
     implementation("io.netty:netty-codec-http")
+    implementation("io.netty:netty-transport-classes-epoll:4.1.90.Final")
```
StefanBratanov: I suppose we don't need the hardcoded version here since artifact is already in `versions.gradle`?

StefanBratanov (approved):
LGTM with one comment

libp2p/build.gradle.kts
ianopolous: Yep not sure why that was there. I've removed it.

libp2p/src/main/java/io/libp2p/core/dsl/HostBuilder.java
```diff
@@ -72,8 +79,20 @@ public Host build() {
...
 
+          secureTransports_.forEach(
+              st ->
+                  b.getSecureTransports()
+                      .add(
+                          p ->
+                              u ->
+                                  st.apply(
+                                      identity.getFactory().invoke(),
+                                      (List<ProtocolBinding<?>>) p)));
```
Nashatyrev: Not sure I understood what is this code for ?

ianopolous: This is to convert the secure transport constructors into normal transport constructors, after passing in the mutable protocols list.

ianopolous: p is the protocol list, u is the connection upgrader (not used), st is the secure transport constructor

ianopolous: The mutable protocols list is necessary so we can add protocols to an already running server. (Which we use to add Identify after getting all listen addresses from a running host)

ianopolous:
To fix the CI android issue I think we need to move to Java 17.

libp2p/src/main/java/io/libp2p/core/dsl/HostBuilder.java
Nashatyrev: Could you please check this PR https://github.com/Peergos/jvm-libp2p/pull/14/files - I've tried to untangle this transport ctor mess a bit

.github/workflows/build.yml
```diff
@@ -11,13 +11,15 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 11
+          java-version: 17
```
Nashatyrev: I would do this in a separate PR if needed 
It makes sense to find out what happened with setup-android github actions (it worked before) 
Probably it sets up the latest SDK when no specific version is specified

ianopolous: Yep, we don't need it, it's just to get CI running.

ianopolous: It turns out we do need java 17 after all to be able to remove bouncy castle TLS provider which otherwise breaks quic.

https://github.com/Peergos/jvm-libp2p/pull/15/files

Thoughts? @Nashatyrev

ianopolous: Actually, by switching to boringssl instead of bouncy castle or jdk I was able to stay on Java 11.

ianopolous:
I think this is good to go now. We now pass interop testing of (tcp, noise, tls, yamux, mplex, quic) to and from rust-libp2p and go-libp2p. We can't talk to zig-libp2p, but that's because they only implement quic and only implement ed25519 TLS signatures. We can add that once netty-quic supports it (https://github.com/netty/netty-incubator-codec-quic/issues/487). 

We had to switch to boringssl for TLSSecureChannel to fix interop and stay on Java 11. But boringssl is faster than JDK TLS provider which itself is much faster than bouncy castle.

libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
```diff
@@ -41,14 +43,26 @@ open class ConnectionOverNetty(
...
+            return toMultiaddr(nettyChannel.localAddress() as InetSocketAddress)
+        }
+    }
+    override fun remoteAddress(): Multiaddr {
+        if (nettyChannel is QuicChannel) {
+            return toMultiaddr(nettyChannel.remoteSocketAddress() as InetSocketAddress)
+        } else {
+            return toMultiaddr(nettyChannel.remoteAddress() as InetSocketAddress)
+        }
+    }
```
Nashatyrev, line 59: This is definitely needs to be refactored, but I would do it in a followup PR

Nashatyrev (approved):
I'm basically fine with the changes (except another technical debt)

libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
Nashatyrev: Here is the supposed refactoring: #408

ianopolous: Looks nice! Want to merge it into this branch (there are 2 other commits here)?

Nashatyrev: I would better isolate it tbh.

ianopolous: Oh, you mean merge this first? Sure.

Nashatyrev:
@tbenr @StefanBratanov do you think this could be merged?

tbenr:
I'll do some tests with teku today

libp2p/src/test/java/io/libp2p/transport/quic/QuicKuboTestJava.java
```diff
@@ -0,0 +1,89 @@
...
+public class QuicKuboTestJava {
+  @Test
+  //    @Disabled
+  void pingKubo() throws Exception {
+    if (System.getProperty("os.name").toLowerCase().startsWith("windows")) return;
+    PeerId peerId = PeerId.fromBase58(getKuboPeerId());
+
+    Host clientHost =
+        new HostBuilder()
+            //                .secureTransport(QuicTransport::Ed25519)
```
tbenr: cleanup here

libp2p/src/test/java/io/libp2p/transport/quic/QuicKuboTestJava.java
```diff
@@ -0,0 +1,89 @@
...
+
+    Assertions.assertThrows(
+        ExecutionException.class, () -> pingCtr.ping().get(5, TimeUnit.SECONDS));
+
+    clientHost.stop().get(5, TimeUnit.SECONDS);
+    System.out.println("Client stopped");
+  }
+
+  private static String getKuboPeerId() throws IOException, URISyntaxException {
+    String url = "http://localhost:5001/api/v0/id";
```
tbenr, line 59: why this is running on windows while the previous isnt?

libp2p/src/test/java/io/libp2p/transport/quic/QuicServerTestJava.java
```diff
@@ -0,0 +1,325 @@
...
+    return new Random().nextInt(20_000) + 10_000;
+  }
+
+  @Test
+  void pingJava() throws Exception {
+    String localListenAddress = "/ip4/127.0.0.1/udp/" + getPort() + "/quic-v1";
+
+    Host clientHost =
+        new HostBuilder()
+            //                .secureTransport(QuicTransport::Ed25519)
```
tbenr: cleanup here

libp2p/src/test/java/io/libp2p/transport/quic/QuicServerTestJava.java
```diff
@@ -0,0 +1,325 @@
...
+            .keyType(KeyType.ED25519)
+            .secureTransport(QuicTransport::Ecdsa)
+            .transport(TcpTransport::new)
+            .secureChannel(TlsSecureChannel::ECDSA)
+            .muxer(StreamMuxerProtocol::getYamux)
+            .build();
+
+    Host serverHost =
+        new HostBuilder()
+            //                .secureTransport(QuicTransport::Ed25519)
```
tbenr: cleanup here

tbenr (approved):
LGTM. just some nits and a question

No regressions on teku

.github/workflows/build.yml
```diff
@@ -12,12 +12,16 @@ jobs:
         with:
           distribution: temurin
           java-version: 11
+      - name: Install and run ipfs
```
tbenr, line 15: I just have a tiny concern here.
We are downloading and running external tool before building the lib. In principle, if that gets compromised it could modify the source and we will build\publish something different.
Maybe I'm paranoid. wdyt @Nashatyrev  ?

install-run-ipfs.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/sh
+wget https://dist.ipfs.io/kubo/v0.34.1/kubo_v0.34.1_linux-amd64.tar.gz -O /tmp/kubo_linux-amd64.tar.gz
```
tbenr, line 2: to solve my concern above, we should add a sha256 check on the downloaded `kubo_v0.34.1_linux-amd64.tar.gz` so we verify that it isn't compromised

tbenr: cc @Nashatyrev

libp2p/src/test/java/io/libp2p/transport/quic/QuicKuboTestJava.java
ianopolous: The install script only handle linux, no point adding windows. So the test is disabled on windows.

install-run-ipfs.sh
ianopolous: This is a good point!

Nashatyrev: Yes, this sounds pretty safe to me

tbenr: i like it now :)

libp2p/src/test/java/io/libp2p/transport/quic/QuicKuboTestJava.java
tbenr: ahh sorry this is not a test, it is just a function.
You can remove the `//    @Disabled` up there.

tbenr:
I pressed "the button" :) thanks a lot @ianopolous, nice work!

Nashatyrev:
Yep, @ianopolous thanks for this and all the previous contributions ❤️

ianopolous:
Thank you for the excellent foundation!

tbenr:
qq @ianopolous @Nashatyrev 

we should advertise this as `quic-v1` in `multiaddr` as per https://github.com/libp2p/docs/blob/master/content/concepts/transports/quic.md#quic-in-libp2p because the underlining quiche is `RCF 9000` compliant. Correct?

so from the ethereum perspective we will have something like:

```
/ip4/0.0.0.0/udp/9001/quic-v1/p2p/16Uiu2HAmAtCoy4N1Th3vJE6fVrEG9BVPLVnaJTDgaf22YowGCZtT
```

ianopolous:
@tbenr Yes that's right.

PR #412, opened 2025-06-10
Use netty core instead of incubator artifact for QUIC

StefanBratanov:
As per https://github.com/netty/netty/pull/14979

Also few migrations needed: https://netty.io/wiki/netty-4.2-migration-guide.html


libp2p/src/main/kotlin/io/libp2p/transport/ws/WebSocketClientInitializer.kt
```diff
@@ -12,12 +12,12 @@ internal class WebSocketClientInitializer(
...
 ) : ChannelInitializer<SocketChannel>() {
 
-    public override fun initChannel(ch: SocketChannel) {
+    override fun initChannel(ch: SocketChannel) {
         val pipeline = ch.pipeline()
 
         pipeline.addLast(HttpClientCodec())
         pipeline.addLast(HttpObjectAggregator(65536))
-        pipeline.addLast(WebSocketClientCompressionHandler.INSTANCE)
+        pipeline.addLast(WebSocketClientCompressionHandler(0))
```
zilm13, line 20: Is it safe to give ulimited buffer here and in `WebSocketServerInitializer.kt`?

zilm13 (approved):
LGTM

libp2p/src/main/kotlin/io/libp2p/transport/ws/WebSocketClientInitializer.kt
StefanBratanov: This retains the same behaviour as before, so decided to not touch it, not sure entirely how safe it is.

PR #413, opened 2025-07-02
[QUIC] Set muxer session and secure session on connection

tbenr:
This is another step forward, but still we don't fully connect peers using teku


StefanBratanov (approved):
LGTM. A good first step.

PR #416, opened 2025-07-11
Apply InboundTrafficLimitHandler only if max bytes > 0

westernbear:
According to [specs](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md#hop-protocol), in circuit relay v2, connection data limit must be applied only if max bytes is bigger than 0.
Therefore InboundTrafficHandler should be added only if resv.maxBytes > 0


StefanBratanov (approved):
LGTM. Thanks for the contribution.

westernbear:
@StefanBratanov Hi, I forgot running spotlessApply so I re-committed after executing spotlessApply
Thank you!

StefanBratanov:
> @StefanBratanov Hi, I forgot running spotlessApply so I re-committed after executing spotlessApply Thank you!

Perfect! Thank you, will wait for the build to complete and will merge it!

PR #424, opened 2025-09-12
Dependencies updates

StefanBratanov:
- hardcoding `ktlint` version
- jmh version is part of the `me.champeau.jmh` plugin so no need to version
- explicitly specify version of `org.junit.platform`


lucassaldanha:
superseded by https://github.com/libp2p/jvm-libp2p/pull/459

PR #425, opened 2025-09-15
Add quic to interop client

lucassaldanha:
- Add support to Quic on interop client
- Remove docker gradle plugin and use static Dockerfile


StefanBratanov (approved):
LGTM

PR #429, opened 2025-09-18
Add websocket to interop client

StefanBratanov:


lucassaldanha:
I ran all interop tests locally using `ws` and they are all working. After we merge this, we need to update the project on test-plans to include the updated commit + add `ws` to the json config file with the protocols.

PR #434, opened 2026-01-21
Updated rpc protobuf definition for partial messages

lucassaldanha:
Updated our protobuf definition of RPC message, adding support for partial messages.

Based on libp2p spec update: https://github.com/libp2p/specs/blob/marco/partial-messages/pubsub/gossipsub/partial-messages.md

And verified on reference go implementation: https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto


StefanBratanov (approved):
LGTM

Issue #441, opened 2026-02-16
Feature flag for supported extensions


lucassaldanha:
Add finer-grained support for enabling/disabling gossipsub extensions. At the moment, the node reports support for all implemented extensions, with no option to disable them.

Ideally, we have a more modular approach that does not require a lot of code to add new extensions.

lucassaldanha:
Revisit this after we finalise the application interface for partial messages.

Issue #436, opened 2026-02-11
Add support for Extension Control Messages (gossipsub v1.3)


lucassaldanha:
Spec: https://github.com/libp2p/specs/blob/marco/partial-messages/pubsub/gossipsub/gossipsub-v1.3.md

- Should be able to handle control extensions when using gossipsub v1.3
- Should send control extensions message when connecting to peers using gossisup v1.3
- ~~Should have a feature flag to have fine-grained control over extensions~~ (https://github.com/libp2p/jvm-libp2p/issues/441)

PR #442, opened 2026-02-16
Sending control extension message to remote peers

lucassaldanha:
The overall logic for sending control extension messages to remote peers is simple:
- always report support for all extensions (finer-grained control to be implemented as part of https://github.com/libp2p/jvm-libp2p/issues/441)
- control extension message is sent whenever a new peer connects or when we are subscribing to a topic (to handle incoming and outgoing connections)
- `GossipExtensionsState` class was added to `GossipRouter` to help keep track of which peers our node has already sent a control extension message to, to avoid sending it again.
- When a peer disconnects, we clear the information associated with that peer from `GossipExtensionsState`
- Control extensions messages are only sent when the gossipsub protocol version supports it (1.3+)

part of #436


StefanBratanov (approved):
LGTM

PR #443, opened 2026-02-17
Gossip Extensions feature flag

lucassaldanha:
**Main change**

The main idea behind this change is to ensure backwards compatibility with anyone already using GossipRouter.

Introduce a new `GossipExtensionConfig` data class that is used to set which extensions should be enabled.

The `GossipRouter` has been updated to receive a new constructor parameter `gossipExtensionsConfig` that is used to initialise the object `GossipExtensionsState`. `GossipRouterBuilder` was updated to receive a `GossipExtensionsConfig`; this way, the application can decide what extensions to enable when creating a router.

When processing extensions messages on `processExtensions`, the router uses the flags from GossipExtentionState to determine if:
1. The local node has a particular extension enabled
2. The peer who sent this message has notified our node that it supports that extension (via ControlExtension message)

**Other changes**
Had to rename a bunch of methods and variables because it wasn't consistent with the spec. We have ControlExtensions (and not ExtensionsControl), etc.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -185,9 +185,7 @@ abstract class AbstractRouter(
             processControl(msg.control, peer)
         }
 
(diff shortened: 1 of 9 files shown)


StefanBratanov (approved):
LGTM

PR #455, opened 2026-04-20
Fix ConnectionClosedException propagating uncaught to Netty event loop

lucassaldanha:
## Summary

- In Netty 4.2.x, writing to a locally-disconnected stream caused a `ConnectionClosedException` to escape `MuxChannel.doWrite()` uncaught, surfacing as spurious `PLEASE FIX OR REPORT` noise in Teku logs.
- The original `throw`+`catch` pattern caught the exception locally, but the subsequent `buf.remove(cause)` call — sitting outside any try/catch — can propagate exceptions back through promise-listener callbacks in certain Netty versions.
- Fix: move the `localDisconnected` check outside the try/catch and wrap `buf.remove(cause)` defensively so nothing can escape `doWrite()`. Pending write promises are still properly failed with `ConnectionClosedException`.

## Test plan

- [x] Added regression test in `MuxHandlerAbstractTest` that sets `localDisconnected = true` while the channel is still active (simulating the window between flag assignment and `deactivate()` in `doDisconnect()`), calls `writeAndFlush`, and asserts the write future is failed with `ConnectionClosedException` — no uncaught exception
- [x] Run `./gradlew :libp2p:test --tests "io.libp2p.mux.*"` — all 34 mux tests pass

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/util/netty/mux/MuxChannel.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/util/netty/mux/MuxChannel.kt
@@ -32,13 +32,20 @@ class MuxChannel<TData>(
         initializer(this)
     }
 
+    @Suppress("SwallowedException")
     override fun doWrite(buf: ChannelOutboundBuffer) {
         while (true) {
             val msg = buf.current() ?: break
+            if (localDisconnected) {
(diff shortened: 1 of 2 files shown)


lucassaldanha:
This should fix https://github.com/Consensys/teku/issues/10201

StefanBratanov (approved):
LGTM, nice fix

PR #472, opened 2026-04-29
Filter backoff peers in subscribe() and make v1.3 honour backoff/IDONTWANT

lucassaldanha:
## Summary

Two correctness fixes in the GossipSub router, both surfaced while debugging a reproducible Teku ↔ Prysm disconnect on Ethereum Gloas/ePBS devnets.

### 1. `GossipRouter.subscribe()` ignored PRUNE backoff (production bug)

`subscribe()` seeded the new mesh from `fanoutPeers` and `otherPeers` filtered by `score >= 0 && !isDirect` only — without the `!isBackOff(peer, topic)` check that all heartbeat-driven mesh maintenance paths apply. When a local `subscribe()` ran shortly after a remote PRUNE (e.g. attestation-subnet rotation right after the remote pruned us), the router immediately GRAFTed the still-backed-off peer.

In go-libp2p-pubsub, every such GRAFT during the backoff window is a P7 behaviour-penalty violation:

```
gossipsub.go:928  GRAFT: ignoring backed off peer <teku-peer>
```

Each violation adds to `behaviourPenalty`. With Prysm's default scoring (`BehaviourPenaltyWeight = -15.92`, `BehaviourPenaltyThreshold = 6`), ~10 violations are enough to push the peer score below Prysm's `-100` disconnect threshold, causing permanent disconnection within minutes. The affected topics in production were `beacon_attestation_0/1/2/3` — exactly the rotating attestation subnets, which is the trigger condition for re-`subscribe()`.

Fix: filter `fanoutPeers` and `otherPeers` by `!isBackOff(it, topic)` in `GossipRouter.subscribe()`, mirroring `JOIN` in the go-libp2p-pubsub reference.

### 2. `supportsBackoffAndPX()` and `supportsIDontWant()` were not cumulative across versions

GossipSub versions are additive: v1.3 ⊃ v1.2 ⊃ v1.1. The exact-match checks meant a router negotiating `/meshsub/1.3.0` silently:
- skipped storing the `Backoff` field from incoming PRUNEs (`GossipRouter.kt:315`)
- omitted `Backoff`/PX from outgoing PRUNEs (`GossipRouter.kt:779`)
- short-circuited `handleIDontWant` (`GossipRouter.kt:367`)

This is the same class of bug as #1, but only triggers once a router (Teku or otherwise) advertises v1.3. Made the support methods cumulative so v1.3 inherits all v1.1/v1.2 behaviour.

## Test plan

- [x] `GossipV1_1Tests.testSubscribeRespectsBackoff` — new regression test. Fails on `develop`: 1 GRAFT sent within the 30 s backoff window. Passes after the `subscribe()` fix.
- [x] `GossipV1_3Tests.testBackoffTimeoutOnV1_3` — new regression test mirroring the existing v1.1 backoff test, exercising a v1.3-negotiated router.
- [x] `PubsubProtocolTest` — pins the cumulative semantics of all three feature-support methods.
- [x] All existing `io.libp2p.pubsub.*` tests pass.
- [x] `./gradlew spotlessApply` and `./gradlew :libp2p:detekt` clean.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubProtocol.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/PubsubProtocol.kt
@@ -19,14 +19,14 @@ enum class PubsubProtocol(val announceStr: ProtocolId) {
      * https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange
      */
     fun supportsBackoffAndPX(): Boolean {
-        return this == Gossip_V_1_1 || this == Gossip_V_1_2
+        return this == Gossip_V_1_1 || this == Gossip_V_1_2 || this == Gossip_V_1_3
     }
 
     /**
      * https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message
      */
     fun supportsIDontWant(): Boolean {
-        return this == Gossip_V_1_2
+        return this == Gossip_V_1_2 || this == Gossip_V_1_3
     }
 
     /**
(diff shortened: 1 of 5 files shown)


StefanBratanov (approved):
LGTM

PR #476, opened 2026-05-13
Addressing gaps identified on QUIC implementation

lucassaldanha:
This PR addresses several correctness issues in the QUIC transport and lays the groundwork for NAT hole punching.

  Bug fixes:
  - Race condition in concurrent inbound connections — replaced the shared Libp2pTrustManager.remoteCert field with per-connection sslEngine().session.peerCertificates, eliminating a data race when multiple peers connect
  simultaneously
  - NPE-prone cert access on dial — replaced the trustManager.remoteCert!! access with the safer sslEngine().session.peerCertificates path; also handles the case where no PeerId is present in the multiaddr by extracting
  it from the TLS certificate post-handshake
  - Spec compliance — disabled unidirectional streams (initialMaxStreamsUnidirectional(0)) since libp2p QUIC exclusively uses bidirectional streams; disabled activeMigration until address-change handling is implemented

  Improvements:
  - QuicConfig — introduces a config data class with sensible defaults (15 MB connection window, 10 MB per-stream window, 256 bidirectional streams, 30 s idle timeout), replacing hardcoded 1 MB / 64-stream values
  - Deterministic stateless reset token — derives a 16-byte token from SHA-256(localKey || "libp2p-quic-stateless-reset") so peers can cleanly close connections after a node restart
  - Port reuse for outbound connections — tracks active listener DatagramChannels by address family; when dialing, attempts to bind the client UDP socket to the same port as the listener (SO_REUSEADDR) to maintain
  consistent NAT mappings

  NAT hole punching foundation:
  - Adds dialAsListener support: sends UDP probe datagrams at 50 ms intervals from the active listener socket to open NAT mappings, then waits up to 5 s for the remote peer to connect back via a pendingHolePunches routing
   map

  Documentation:
  - Documents that private networks (PSK) are incompatible with the QUIC transport due to QUIC's mandatory TLS 1.3

  ## Test plan
  - concurrentInboundConnections — verifies no race condition under concurrent inbound connections
  - dialWithoutPeerIdExtractsPeerIdFromCert — verifies PeerId is correctly extracted from TLS cert when not present in the multiaddr
  - holePunchTimeout — verifies the hole punch path times out cleanly after 5 s with no connection
  - Full QUIC test suite passes: ./gradlew :libp2p:test --tests "io.libp2p.transport.quic.*"

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicConfig.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicConfig.kt
@@ -1,0 +1,12 @@
+package io.libp2p.transport.quic
+
+import java.time.Duration
+
+data class QuicConfig(
+    val connectTimeout: Duration = Duration.ofSeconds(15),
+    val idleTimeout: Duration = Duration.ofSeconds(30),
+    val maxConnectionData: Long = 15L * 1024 * 1024,
+    val maxStreamDataLocal: Long = 10L * 1024 * 1024,
+    val maxStreamDataRemote: Long = 10L * 1024 * 1024,
+    val maxStreamsBidirectional: Long = 256,
+)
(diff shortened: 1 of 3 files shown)


StefanBratanov (approved):
LGTM

PR #477, opened 2026-05-19
Add PubsubRpcLimits decoding count-cap to reduce memory footprint

lucassaldanha:


libp2p/src/main/kotlin/io/libp2p/pubsub/RpcCountFrameDecoder.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * When [limits] has every count null and `rejectEmptyPublishEntries == false`
+ * (e.g. [PubsubRpcLimits.NONE]) the validator walks the buffer once and accepts
+ * every well-formed frame — there is no special-cased fast-path.
+ */
+class RpcCountFrameDecoder(private val limits: PubsubRpcLimits) : MessageToMessageDecoder<ByteBuf>() {
+
+    override fun decode(ctx: ChannelHandlerContext, msg: ByteBuf, out: MutableList<Any>) {
+        val result = try {
+            RpcMessageCountValidator.validate(msg, limits)
```
StefanBratanov, line 37: Isn't it better when PubsubRpcLimits is NONE we don't walk the buffer at all, it will fail when converting to the objecy anyways in this case

StefanBratanov (approved):
LGTM, apart from one thing

libp2p/src/main/kotlin/io/libp2p/pubsub/RpcCountFrameDecoder.kt
lucassaldanha: True. I'll add it.

PR #479, opened 2026-05-21
Release ByteBuf on control frames and invalid stream tag

lucassaldanha:
## Summary
- `MplexHandler.channelRead()` now releases the frame buffer for OPEN/CLOSE/RESET. Previously only DATA frames forwarded the buffer to a child channel, so the retained slice created by `MplexFrameCodec.decode()` leaked on every control frame.
- `MplexFrameCodec.decode()` now releases the retained data slice when the stream tag lookup throws. With `streamTag == 7` (invalid), `MplexFlag.getByValue()` threw after `data.retain()`, leaking the underlying inbound buffer.

## Test plan
- [x] New unit tests reproduce the leaks before the fixes and pass after:
  - `MplexHandlerTest`: OPEN/CLOSE/RESET frame data buffer is released, repeated OPEN frames do not leak.
  - `MplexFrameCodecTest`: invalid stream tag does not leak inbound buffer.
- [x] `./gradlew :libp2p:test --tests 'io.libp2p.mux.*' :libp2p:spotlessCheck` passes.


StefanBratanov (approved):
LGTM

PR #480, opened 2026-05-21
Verify message `from` matches signing key (01)

lucassaldanha:
Summary:

* PubsubCrypto.pubsubValidate now rejects messages missing/empty `from`, `key` or `signature`, returns false (never throws) on un-parseable key/from, and requires `PeerId.fromPubKey(msg.key) == PeerId(msg.from)` in addition to the existing signature check.

* PubsubApiImpl.PublisherImpl.publishExt rejects any caller-supplied `from` that does not match the publisher's signing key, throwing IllegalArgumentException. Null `from` continues to default to the signing key's PeerId as before; unsigned publishers (no privKey) are unchanged.

* AbstractRouter no longer leaves a seenMessages entry behind for a message that fails validation. On both sync and async validation failure the entry inserted speculatively at the start of `onInbound` is evicted, so a later legitimate message with the same id is still processed and delivered.

Adds PubsubStrictSignSpoofTest covering forged-from rejection, missing-field rejection, publishExt mismatch rejection, the matching-from happy path, and an end-to-end regression test that the seen-cache is no longer poisoned by a rejected forgery.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -226,10 +226,8 @@ abstract class AbstractRouter(
                 true
             } catch (e: Exception) {
                 logger.debug("Invalid pubsub message from peer {}: {}", peer, it, e)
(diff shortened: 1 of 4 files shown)


StefanBratanov (approved):
LGTM

PR #487, opened 2026-06-10
fix(quic): dial from ephemeral port instead of reusing listener port

lucassaldanha:
When the dialing transport had an active QUIC listener, dial() tried to bind a separate NioDatagramChannel to the listener's own port using only SO_REUSEADDR. On Linux two UDP sockets cannot share a port without SO_REUSEPORT (an Epoll-only option), so the bind threw BindException. The fallback then re-used the same non-@Sharable QUIC codec handler instance on a second channel, which also throws, so the outbound connection never completed and timed out — 0 successful outbound dials on listening nodes.

Always bind the dial socket to an ephemeral port (the path already exercised by every passing QUIC test). NAT-consistent dialing for hole punching is unaffected: dialAsListener reuses the listener socket directly.

Add QuicServerTestJava.dialWhileListeningCompletes regression test that dials while a listener is active — the previously-untested broken branch.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
@@ -215,61 +215,36 @@ class QuicTransport(
             .build()
 
         val targetAddr = fromMultiaddr(addr) as InetSocketAddress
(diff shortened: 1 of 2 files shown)


StefanBratanov (approved):
LGTM

PR #489, opened 2026-06-14
Fix quic close write logic

lucassaldanha:
Two fixes to bring QUIC streams handling closer to how we handle when using the muxer on TCP.

 1. Close quietly on remote STREAM_RESET — when a peer resets a stream (happens a lot on disconnects and cancelled RPCs), avoid Netty firing QuicStreamResetException and leaving the stream open.
  2. Make closeWrite accept already-closed streams beacuse of how Teku's Goodbye flow closes the connection (and its streams).


rolfyone (approved):
LGTM

PR #490, opened 2026-06-14
fix(quic): close leaked dial datagram channels; harden NoiseXXCodec OOM handling

lucassaldanha:
## Overview

Two fixes that came out of investigating an `OutOfMemoryError` reported on a soak node. The investigation surfaced two distinct defects, so this PR addresses both:

1. **The actual leak was off-heap, in the QUIC transport:** every outbound QUIC dial leaked its ephemeral `NioDatagramChannel`, growing direct memory ~3 GiB/day under connection churn.
2. **`NoiseXXCodec` was masking failures:** it swallowed fatal JVM `Error`s (incl. `OutOfMemoryError`) instead of failing fast, which is how the symptom first showed up.

JVM **heap** usage was flat the whole time — the growth was in `jvm_buffer_pool_used_bytes{pool="direct"}` / process RSS. A heap-dump reverse-reference scan confirmed the cause: ~1,085 orphaned `NioDatagramChannel`s (each holding a codec, native quiche config and a `PendingWriteQueue` backlog of direct buffers) against only ~206 live QUIC connections.

---

## Fix 1 — QUIC: close the ephemeral dial datagram channel (`QuicTransport`)

Each outbound `dial()` binds its own ephemeral `NioDatagramChannel` to carry the QUIC connection. Closing a `QuicChannel` does **not** close its parent datagram channel in Netty, so when the connection ended (reset / timeout / normal close) the datagram channel — plus its codec, native quiche config, and queued direct buffers — was leaked for the lifetime of the transport. Under normal connection churn this is a linear off-heap leak.

This change:
- **Ties the datagram channel lifecycle to the `QuicChannel`:** closes it on `QuicChannel` close, and on connect failure.
- **Closes the established `QuicChannel` when post-handshake setup fails** (e.g. the connection handler throws), so a failed dial can't strand the connection and its datagram channel.

> Scope note: this is the leak fix. Sharing the client `QuicSslContext` across dials (an efficiency + PeerId-validation change) and broader error-path hardening are intentionally **deferred to follow-up PRs**, since the former touches handshake security and deserves its own review.

## Fix 2 — Noise: stop `NoiseXXCodec` swallowing fatal errors

Previously, the `else` branch in `NoiseXXCodec.exceptionCaught` logged any unexpected throwable (including `OutOfMemoryError`) at ERROR but **left the channel open and did not rethrow**, masking the failure and letting the Netty event-loop thread keep running in a degraded state.

This change:
- Closes the channel on **any** unexpected throwable (previously the connection was leaked).
- Adds a dedicated branch for fatal JVM `Error`s that closes the channel **and rethrows**, so the failure is surfaced rather than downgraded to a single log line.

> Note: this is defensive hardening — it stops the failure being masked and the connection leaking; it does not by itself prevent heap exhaustion. The OOM that triggered the investigation was off-heap (Fix 1), not heap.

## Tests

**QUIC (`QuicServerTestJava`, TDD — watched to fail first):**
- the underlying datagram channel is closed when the dial's connection closes
- a dial that fails after the handshake does not leak a connection

**Noise (`NoiseXXCodecTest`, TDD — watched to fail first):**
- fatal `Error` is rethrown and not swallowed
- fatal `Error` closes the channel
- unexpected `RuntimeException` closes the channel

`./gradlew :libp2p:test` (QUIC + Noise suites), `:libp2p:spotlessCheck`, and `:libp2p:detekt` all pass.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/types/OtherExt.kt
@@ -44,3 +44,9 @@ fun <T : Throwable> Throwable.hasCauseOfType(clazz: KClass<T>) =
     Throwables.getCausalChain(this)
         .filter(clazz::isInstance)
         .any()
+
+fun <T : Throwable> Throwable.findCauseOfType(clazz: KClass<T>): T? =
+    Throwables.getCausalChain(this)
+        .filter(clazz::isInstance)
+        .map { clazz.java.cast(it) }
+        .firstOrNull()
--- a/libp2p/src/main/kotlin/io/libp2p/security/noise/NoiseXXCodec.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/security/noise/NoiseXXCodec.kt
@@ -49,8 +49,16 @@ class NoiseXXCodec(val aliceCipher: CipherState, val bobCipher: CipherState) :
         } else if (cause.hasCauseOfType(SecureChannelError::class)) {
             logger.debug("Invalid Noise content", cause)
             closeAbruptly(ctx)
+        } else if (cause.hasCauseOfType(Error::class)) {
+            // A fatal JVM error (e.g. OutOfMemoryError) must never be swallowed: close the
+            // channel to release resources, then rethrow so the failure is surfaced rather
+            // than downgraded to a single log line while the event loop keeps running.
+            logger.error("Fatal error in Noise channel, closing", cause)
+            closeAbruptly(ctx)
+            throw cause
         } else {
             logger.error("Unexpected error in Noise channel", cause)
+            closeAbruptly(ctx)
         }
     }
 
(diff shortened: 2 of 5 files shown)


rolfyone:
review comments follow:
[P2] QuicTransport.kt (line 260): cancelled QUIC dials can still leave the upstream QuicChannel running. NetworkImpl.connect() cancels losing dial futures, and TCP explicitly closes its channel on cancellation, but this QUIC dial() returns only the dependent thenApply future. Cancelling that future does not cancel/close quicConnFuture; if the QUIC connect later succeeds, the thenApply body is skipped, the channel is not registered or handed to the caller, and the new UDP-parent close listener only helps once something else closes the child. This leaves a resource leak or at least delayed cleanup in the normal multi-address race path.

[P2] NoiseXXCodec.kt (line 52): wrapped fatal Errors are detected but not rethrown as the fatal error. The branch uses hasCauseOfType(Error::class) but then throw cause; for outbound encode() failures Netty can wrap Throwable in EncoderException, so an OutOfMemoryError from the allocation in encode() would be rethrown as a runtime wrapper. That still downgrades the fatal JVM signal. I’d extract and throw the actual Error from the causal chain, plus add a test with EncoderException(OutOfMemoryError(...)).

rolfyone (approved):
LGTM

PR #492, opened 2026-06-17
fix(tls,quic): expose libp2p host pubkey as remotePubKey, not the ephemeral cert subject key 2/2

lucassaldanha:
## Summary

This PR contains two related fixes to the QUIC transport.

### 1. Expose the libp2p host pubkey as `remotePubKey` (hole-punch path)

Part 2/2 of exposing the libp2p host public key (not the ephemeral X.509 cert
subject key) as `SecureChannel.Session.remotePubKey`. Part 1/2 (the TLS channel
and the QUIC server-inbound / client-dial paths) is already merged into
`develop`; this completes the QUIC transport by bringing the **hole-punch /
direct connection path** in line with the rest.

This upholds the invariant `PeerId.fromPubKey(remotePubKey) == remoteId` that
SecIO, Noise, and Plaintext already guarantee and that downstream consumers
rely on (e.g. signed-record verification).

### 2. Cache connection addresses to avoid NPE on a freed QUIC channel

`ConnectionOverNetty` resolved local/remote addresses from the underlying Netty
channel on every call. For QUIC the native channel state is freed on close,
after which `remoteSocketAddress()` returns `null` and the `!!` in
`QuicTransport.remoteAddress()` throws a `NullPointerException`. This surfaced
when a disconnect handler read the remote address during idle-timeout teardown
(observed downstream in Teku's `LibP2PPeer.internalDisconnectImmediately`).

## Changes

`QuicTransport.kt` (hole-punch path):

- Both `remoteId` and `remotePubKey` for the `SecureChannel.Session` now come
  from the single libp2p host key extracted from the cert extension
  (`verifyAndExtractIdentity`), never the ephemeral cert subject key.
- Add a defence-in-depth check: if a dial-target peer id is known, assert the
  libp2p key extracted from the cert hashes to it, throwing `Libp2pException`
  otherwise. This mirrors the guard already on the client-dial path and means a
  misconfigured trust manager cannot silently leak the wrong identity into the
  connection.
- Remove the now-unused `remotePeerId` / `remotePubKey` locals.

`ConnectionOverNetty.kt`:

- Cache the local and remote `Multiaddr` on first successful read while the
  channel is still live, so teardown-time callers get a stable value instead of
  dereferencing the freed native channel.

## Testing

- `QuicRemotePubKeyIdentityTest` (added with part 1/2, already on `develop`)
  exercises a real QUIC handshake between two hosts and asserts
  `PeerId.fromPubKey(secureSession.remotePubKey) == secureSession.remoteId` on
  both the client and server connections.
- `ConnectionOverNettyTest` (new) verifies that both `localAddress()` and
  `remoteAddress()` return the cached value and do not re-hit the transport
  after the underlying channel has been freed.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
@@ -50,4 +50,15 @@ open class ConnectionOverNetty(
 
     override fun remoteAddress(): Multiaddr =
         cachedRemoteAddress ?: nettyTransport.remoteAddress(nettyChannel).also { cachedRemoteAddress = it }
+
+    /**
+     * Eagerly resolves and caches the local and remote addresses while the underlying channel is
+     * still live. Transports whose channels free their native state on close (e.g. QUIC) must call
+     * this before exposing the connection, so a teardown-time caller that is the first to read an
+     * address gets the cached value instead of dereferencing the freed channel.
+     */
+    fun cacheAddresses() {
+        localAddress()
+        remoteAddress()
+    }
 }
--- a/libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt
@@ -20,6 +20,16 @@ open class ConnectionOverNetty(
     private lateinit var muxerSession: StreamMuxer.Session
     private lateinit var secureSession: SecureChannel.Session
 
+    // Addresses are resolved from the underlying Netty channel, but some transports (e.g. QUIC)
+    // free the native channel state on close, after which the socket addresses are no longer
+    // available. We cache the addresses on first successful read while the channel is still live
+    // so that teardown-time callers (e.g. disconnect handlers) get a stable value instead of an NPE.
+    @Volatile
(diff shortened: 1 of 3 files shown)


rolfyone:
[P2] [ConnectionOverNetty.kt (line 48)](/private/tmp/jvm-libp2p-pr492/libp2p/src/main/kotlin/io/libp2p/transport/implementation/ConnectionOverNetty.kt:48) still only caches after a successful first address read. If the disconnect/teardown handler is the first caller after QUIC has freed native state, it still falls through to QuicTransport.remoteAddress() and can hit the same remoteSocketAddress()!! NPE. NetworkImpl does not warm this cache when registering connections, and the new test only covers “read once while live, then read after freed.” I’d warm local/remote addresses before exposing the connection, or pass pre-resolved addresses into the connection for QUIC.
[P2] [QuicTransport.kt (line 539)](/private/tmp/jvm-libp2p-pr492/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt:539) throws on hole-punch peer-id mismatch after registerChannel(quicChannel) at line 526. Unlike the normal dial path, this thenApply has no catch/close cleanup, so a mismatched or misconfigured peer can leave an accepted QUIC channel registered and open while the returned future fails. Wrap the hole-punch setup in the same close-on-failure pattern as the normal dial path.

rolfyone (approved):
LGTM

PR #493, opened 2026-06-18
Gossip heartbeat allocations reduction

tbenr:
# Gossipsub Allocation Optimizations

Two allocation hotspots in the gossip heartbeat, found by profiling Teku (mainnet head,
all-subnets, `async-profiler -e alloc`) and fixed in this repo.

### 1. Gossip message-ids — build them once per heartbeat instead of per topic
`MCache.getMessageIds(topic)` was called once per topic from `emitGossip()`, which runs only
inside `heartbeat()`. Its `history.takeLast(gossipSize).flatten()` is **topic-independent**, yet
it was rebuilt for every topic — ~64+ times per heartbeat under all-subnets — each time growing
a default-capacity `ArrayList` (`grow → Arrays.copyOf`) and then re-filtering/dedup per topic.

Fix: replace it with `MCache.getGossipMessageIdsByTopic(): Map<Topic, Set<MessageId>>` that makes
**one pass** over the last `gossipSize` history windows (via `listIterator(size - gossipSize)`,
no intermediate `takeLast`/`flatten` lists) and buckets ids into per-topic `LinkedHashSet`s
(dedup + insertion order preserved). `heartbeat()` calls it **once** and passes the map to each
`emitGossip(topic, peers, map)`, which just does `map[topic] ?: return`.

This is stateless — a pure function of `history`, no cache and no invalidation contract — and
removes both the per-topic re-flatten **and** the per-topic dedup rescan.

### 2. `DefaultGossipScore.score` — drop the boxing in the topic-score sum
`peerScore.topicScores.values.map { it.calcTopicScore() }.sum()` allocates an intermediate
`List<Double>` and **boxes one `Double` per (peer, topic)** on every `score()` call (called
O(peers × topics) times per heartbeat).

Fix: `…values.sumOf { it.calcTopicScore() }` — folds with a primitive `double` accumulator;
no list, no boxing. (The remaining `CappedValueDelegate` boxing is structural to the generic
delegate and left untouched.)

## Impact (Teku, mainnet head, `--p2p-subscribe-all-subnets-enabled`)

Measured in **bytes** with `async-profiler -e alloc --total`, two 300s steady-state windows,
baseline `1.3.2-RELEASE` vs the patched build:

| Path | Baseline 1.3.2 | Patched | Saved / epoch |
|---|---|---|---|
| msg-id build (`getMessageIds` → `getGossipMessageIdsByTopic`) | 16.6 MB/s · 6.2 GB/epoch · 3.19% | 0.35 MB/s · 132 MB/epoch · 0.08% | −6.1 GB |
| `DefaultGossipScore.score` | 13.5 MB/s · 5.1 GB/epoch · 2.60% | 0.56 MB/s · 211 MB/epoch · 0.13% | −4.9 GB |
| **Combined** | **30.1 MB/s · 11.3 GB/epoch · 5.79%** | **0.92 MB/s · 344 MB/epoch · 0.22%** | **−10.9 GB** |

**Combined saving ≈ 29 MB/s ≈ 10.9 GB/epoch ≈ ~2.4 TB/day — a 97% reduction** of these two
paths' allocation. (`%` = share of all allocation in each run.)

> ⚠️ These figures are for a **super-node meshed on ~64+ topics** (all attestation + custody
> subnets). The savings scale with topic × peer count; a default-subscription node (\~2 subnets)
> would see a small fraction of the absolute numbers. `--total` is sampled (±\~10%), and the two
> windows are different mainnet traffic, so only the per-path delta is attributable to the fix.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt
@@ -668,6 +668,8 @@ open class GossipRouter(
         }
 
         try {
+            val gossipMessageIdsByTopic = mCache.getGossipMessageIdsByTopic()
+
             mesh.entries.forEach { (topic, peers) ->
 
                 // drop underscored peers from mesh
@@ -718,7 +720,7 @@ open class GossipRouter(
                     }
                 }
 
-                emitGossip(topic, peers)
+                emitGossip(topic, peers, gossipMessageIdsByTopic)
             }
             fanout.entries.forEach { (topic, peers) ->
                 peers.removeIf {
@@ -731,7 +733,7 @@ open class GossipRouter(
                         .shuffled(random)
                         .take(needMore)
                 }
-                emitGossip(topic, peers)
+                emitGossip(topic, peers, gossipMessageIdsByTopic)
             }
             lastPublished.entries.removeIf { (topic, lastPub) ->
                 (currentTimeSupplier() - lastPub > params.fanoutTTL.toMillis())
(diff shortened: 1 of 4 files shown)


Nashatyrev (approved):
LGTM 👍

PR #494, opened 2026-06-18
fix(quic): validate hole-punch peer identity early and handle STOP_SENDING

lucassaldanha:
## Summary

Two follow-up QUIC transport fixes on top of #492 (which is merged into `develop`).

### 1. Validate hole-punch peer identity before exposing the connection

On the hole-punch inbound path, `channelActive` set the secure/muxer sessions and
completed the pending future with the raw `QuicChannel` **before** the dial-target peer
id was checked — that check only ran later in `dialAsListener`'s `thenApply`. Because the
server codec installs an `InboundStreamHandler`, a peer that reached the expected UDP
tuple but presented a valid certificate for the **wrong** libp2p identity could have
inbound streams dispatched to protocol handlers during the window before the later check
closed the channel (an unauthenticated-by-target window into inbound protocols).

The dial-target validation now happens in `channelActive`, before the connection is
exposed:

- Each pending hole punch carries its expected peer id (`PendingHolePunch`).
- `holePunchIdentityMatches(expected, actual)` is extracted and unit-tested (null expected
  ⇒ accept any, matching a normal inbound server connection).
- On mismatch, the channel is closed and the future failed **before** the secure session is
  set or the future completed. `channelActive` and stream-activation events run serially on
  the channel event loop, so closing synchronously here closes the window.
- The existing `dialAsListener` check is retained as defence-in-depth.

### 2. Swallow STOP_SENDING `ChannelOutputShutdownException` on streams

When a remote peer sends a STOP_SENDING frame (it no longer wants to read from a stream)
while the local side has queued writes, Netty fails those writes with
`ChannelOutputShutdownException("STOP_SENDING frame received")` and surfaces it into the
stream channel's pipeline. `QuicStreamReadCloseEventConverter` only special-cased
`QuicStreamResetException`, so this exception was forwarded to application handlers and
logged as an unhandled error (e.g. Teku "Unhandled error while processing req/response").

STOP_SENDING is a benign half-close: only our write side is affected, the read side may
still deliver data. The converter's `exceptionCaught` now swallows
`ChannelOutputShutdownException` quietly **without** closing the stream (matching Netty's
own "should not close the channel" handling of STOP_SENDING), mirroring the existing
remote-reset handling.

## Testing

- `QuicHolePunchIdentityTest` — unit tests for the `holePunchIdentityMatches` gate
  (match / mismatch / null-expected).
- `QuicStreamReadCloseEventConverterTest` — deterministic `EmbeddedChannel` unit tests:
  STOP_SENDING swallowed without surfacing or closing, remote-reset closes quietly, and
  unrelated exceptions still propagate.
- Existing `QuicServerTestJava` integration suite (including the remote-reset and
  closeWrite tests) still passes; `spotlessCheck` clean.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicStreamReadCloseEventConverter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicStreamReadCloseEventConverter.kt
@@ -12,6 +12,13 @@ import io.netty.handler.codec.quic.QuicStreamResetException
  */
 class QuicStreamReadCloseEventConverter : ChannelInboundHandlerAdapter() {
 
+    companion object {
+        // Netty's quiche codec (QuicheQuicStreamChannel) raises ChannelOutputShutdownException with
+        // exactly this message when the remote sent STOP_SENDING. The only other message it uses is
+        // "Fin was sent already" for a local write-after-FIN, which must NOT be swallowed.
+        private const val STOP_SENDING_MESSAGE = "STOP_SENDING frame received"
+    }
+
     override fun userEventTriggered(ctx: ChannelHandlerContext, evt: Any) {
         if (evt == ChannelInputShutdownReadComplete.INSTANCE) {
             ctx.fireUserEventTriggered(RemoteWriteClosed)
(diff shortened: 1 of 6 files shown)


lucassaldanha:
This PR blocks https://github.com/Consensys/teku/issues/10866

rolfyone:
Residual risk: the hole-punch identity guard has helper-level tests, but not an end-to-end wrong-identity hole-punch regression test.

want to add an extra test case?

rolfyone (approved):
LGTM

PR #504, opened 2026-07-15
QUIC memory leak fix

tbenr:
## Summary

Fix unbounded outbound pubsub writes when a remote peer stops consuming a QUIC stream.

A stalled peer could leave millions of unresolved writes retained by
`QuicheQuicStreamChannel`, causing extreme heap growth. The fix integrates Netty
channel writability with the pubsub outbound pump and bounds retained work per
peer.

## What changed

- Gate outbound pubsub materialization on the outbound stream's `isWritable()`
  state.
- Allow only one unresolved transport write per peer's outbound pubsub stream.
- Resume the pump on `channelWritabilityChanged` when QUIC flow control opens.
- Serialize queued RPC batches while preserving ordering.
- Enforce per-peer retained byte and logical-entry limits.
- Use bounded defaults:
  - bytes: `maxOf(8 MiB, maxGossipMessageSize + 4 MiB)`
  - entries: `10,000`
- Enforce the existing progress deadline when a peer makes no write progress.
  Successful write completion resets the deadline; writability changes alone do
  not.
- On write failure, progress timeout, or queue overflow:
  - fail active and queued publication futures;
  - release per-peer accounting;
  - clear outbound state;
  - reset the outbound pubsub stream.
- Keep reset-on-overflow as the initial bounded policy. Selective message
  eviction for slow peers can be considered separately.

## Verification

- Deterministic tests cover writability gating, queue accounting, overflow,
  write sequencing, deadlines, cleanup, and reconnects.
- Real QUIC coverage verifies flow-control pause and resume, at most one
  unresolved write, completion of queued publications, and connection liveness.
- TCP/Mplex coverage verifies healthy publication ordering and batching remain
  unchanged.
- `spotlessCheck` and `detekt` pass.

The original leak involved approximately 4.1 million pending writes, 1.5 GB of
pending data, and 2.9 GB of heap while QUIC send capacity was exhausted.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/util/P2PService.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/util/P2PService.kt
@@ -90,6 +90,17 @@ abstract class P2PService(
                 streamActive(this)
             }
         }
+
+        override fun channelWritabilityChanged(ctx: ChannelHandlerContext) {
+            ctx.fireChannelWritabilityChanged()
+            val peer = peerHandler ?: return
+            runOnEventThread(peer) {
+                if (!aborted && !closed) {
+                    streamWritabilityChanged(this)
+                }
+            }
+        }
+
         override fun channelUnregistered(ctx: ChannelHandlerContext?) {
             closed = true
             runOnEventThread(peerHandler) {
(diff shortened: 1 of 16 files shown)


Nashatyrev:
TBH, I would better use `isWritable()` channel attribute which is basically intended for backpressure. I actually have a commit which introduced this: https://github.com/Nashatyrev/jvm-libp2p/commit/e187d9d4112765f2c12680a175331eb40cac583b

The approach in my commit is slightly different: we are submitting lazy computable writes, i.e. the next write is computed only when writability allows. This may benefit when a message is rebroabcasted: until backpressure allows the message payload is not serialized for every mesh peer and only one instance of it retained. 

Another concern is that 'slow' remote peer still might not trigger stream abortion, but the messages might now be buffered uncapped in the `RpcPartsQueue`. Not sure what's the right strategy to address this TBH. One option is to limit the `RpcPartsQueue` and drop the peer which causes it to exceed. Another option is just to evict older messages in the queue.

tbenr:
closing in favor of the alternative

PR #505, opened 2026-07-23
Support backpressure in P2PService (Gossipsub base class)

Nashatyrev:
This is alternative to #504 with narrower scope. 

This PR extracts generic outbound backpressure support into `P2PService`.

The main change is the new `BackpressureAwarePump`, which serializes outbound polling/writes across threads and pauses polling while the peer outbound channel is not writable. `P2PService` now exposes this through `notifyOutboundDataAvailable(peer)` and `pollOutboundMessage(peer)`.

Pubsub routers were wired through this new mechanism, and shared router tests now cover the basic backpressure behavior for both Flood and Gossip.

Note: the Gossip backpressure integration is intentionally very naive. It only stops polling new outbound RPCs while the channel is not writable; it does not provide bounded queues or prevent router-side buffer growth/overflow under sustained pressure.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/util/BackpressureAwarePump.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/util/BackpressureAwarePump.kt
@@ -3,11 +3,28 @@ package io.libp2p.etc.util
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 
+/**
+ * Outbound message together with the promise that should be completed when the write finishes.
+ */
 data class MessageAndPromise(
     val messagePayload: Any,
     val writePromise: CompletableFuture<Unit>,
 )
 
+/**
+ * Drains outbound messages while respecting channel backpressure.
+ *
+ * The pump is driven by two external signals: [onNewOutboundData] tells it that the supplier may
+ * now have another message, and [onChannelWritabilityChanged] tells it whether polling new messages
+ * is currently allowed. Both methods are synchronized and may be called from different threads.
+ *
+ * [messageSupplier] returns the next already-consumed outbound message, or `null` when no message is
+ * ready. Once a non-null message is returned, the pump writes it even if channel writability changes
+ * to false before the supplier future completes. Backpressure only prevents polling the next message.
+ *
+ * [messageWriter] writes one supplied message and completes with the channel writability observed
+ * after that write was scheduled.
+ */
 class BackpressureAwarePump(
     private val messageWriter: BackpressureAwareAsyncWriter,
     private val messageSupplier: () -> CompletionStage<MessageAndPromise?>
@@ -23,7 +40,7 @@ class BackpressureAwarePump(
     private var outboundDataGeneration = 0L
 
     /**
-     * Notified when channel writability changes
+     * Notifies the pump that channel writability changed.
      */
     @Synchronized
     fun onChannelWritabilityChanged(isWritable: Boolean) {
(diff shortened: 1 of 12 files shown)


libp2p/src/main/kotlin/io/libp2p/etc/util/BackpressureAwarePump.kt
```diff
@@ -0,0 +1,114 @@
...
+     * Notifies the pump that outbound data may now be available from [messageSupplier].
+     */
+    @Synchronized
+    fun onNewOutboundData() {
+        outboundDataGeneration++
+        if (state == State.IDLE && channelWritable) {
+            takeNextMessage()
+        } else if (state == State.IDLE) {
+            state = State.BLOCKED
+        }
```
tbenr: maybe a nested if ?

tbenr: or use `when` for consistency

tbenr:
**Findings**
1. **[P1] Writer failures permanently wedge the pump.** `BackpressureAwarePump.kt:92 types the exceptional stage’s null result as non-null `Boolean`, so `onMessageWritten` is never called and state remains `RUNNING`. Reproduced with a regression test.

2. **[P1] Disconnecting drops queued write promises.** `AbstractRouter.kt:314` removes queued `MessageAndPromise` objects without failing their promises. The caller’s `publish()` future then hangs forever. Reproduced for Floodsub and Gossipsub.

3. **[P1] Event-loop rejection leaves a consumed message unresolved.** `BackpressureAwareAsyncWriter.kt:32` does not catch `execute()` rejection during channel shutdown, leaving both futures and the pump stuck.

4. **[P2] Empty queues throw instead of returning `null`.** `AbstractRouter.kt:344` uses `removeFirst()`, although an empty queue is the pump’s normal termination condition. `pollFirst()` matches the declared contract.

BackpressureAwarePumpTest
```kotlin
    @Test
    fun `recovers after writer failure`() {
        val suppliedMessages = ConcurrentLinkedQueue<CompletableFuture<MessageAndPromise?>>()
        val writer =
            BackpressureAwarePump(
                { _, _ -> CompletableFuture.failedFuture(IllegalStateException("write failed")) },
                {
                    CompletableFuture<MessageAndPromise?>().also(suppliedMessages::add)
                }
            )

        writer.onNewOutboundData()
        suppliedMessages.remove().complete(message("first"))
        writer.onNewOutboundData()

        assertThat(suppliedMessages).hasSize(1)
    }
```

PubsubRouterTest
```kotlin
    @Test
    fun `queued publish completes when peer disconnects`() {
        val fuzz = DeterministicFuzz()

        val router1 = fuzz.createTestRouter(routerFactory)
        val router2 = fuzz.createTestRouter(routerFactory)
        router2.router.subscribe("topic1")

        val connection = router1.connectSemiDuplex(router2, LogLevel.ERROR, LogLevel.ERROR)
        val router1OutboundChannel = connection.conn1.ch1
        router1OutboundChannel.setWritableForTest(false)

        val msg = newMessage("topic1", 2L, "Hello".toByteArray())
        val publishFut = router1.router.publish(msg)
        router1OutboundChannel.runPendingTasks()
        assertThat(publishFut).isNotDone()

        connection.disconnect()
        router1OutboundChannel.runPendingTasks()

        assertThat(publishFut).isDone()
    }
```

Nashatyrev:
> Findings

Oh geese, all 4 are good catches 👍

Nashatyrev:
All resolved

tbenr (approved):
LGTM

lucassaldanha (approved):

PR #506, opened 2026-07-27
Improve Gossip backpressure 

Nashatyrev:
## Summary

Follow-up to #505.

This PR improves pubsub backpressure handling by delaying intermediary `Rpc.RPC` construction until the outbound channel is actually writable. Until then, pending outbound data is stored as RPC parts rather than fully materialized RPC messages.

This lets message payload instances be shared across outbound queues, which may significantly reduces memory pressure. This was also visible in recent simulation experiments.

## Details

- Refactored `PendingRpcParts` / `RpcPartsQueue` so queued RPC parts also own and propagate write promises.
- Changed outbound polling so a ready-to-send RPC batch is generated synchronously only when the channel can write.
- Added `submitPublishMessageSilently()` for forwarded messages that do not need a user-facing publish promise.
    - This avoids allocating promises for internal message forwarding.
    - The memory reduction was significant in attestation propagation simulations.

## Notes

`IDONTWANT` messages are still written directly without going through the backpressure-aware queue. This is probably acceptable as-is, but we may later consider prioritized outbound queues if we want all pubsub control traffic to participate in the same backpressure model.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/core/Libp2pException.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/core/Libp2pException.kt
@@ -71,4 +71,8 @@ open class ProtocolViolationException(message: String) : Libp2pException(message
  * When trying to write a message to a peer within [io.libp2p.etc.util.P2PServiceSemiDuplex]
  * but there is no yet outbound stream created.
  */
-open class SemiDuplexNoOutboundStreamException(message: String) : Libp2pException(message)
+open class SemiDuplexNoOutboundStreamException(message: String = "SDPeerHandler.otherStreamHandler hasn't been initialized") :
+    Libp2pException(message)
+
+open class StreamNotActiveException(message: String = "Stream is either closed already or not activated yet") :
+    Libp2pException(message)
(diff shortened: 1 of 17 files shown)


tbenr:
meantime gpt:

**Findings**
1. **[P1] Unbounded batching can exceed `maxGossipMessageSize`.** While blocked, flush boundaries disappear. With default `maxPublishedMessages = null`, all queued publishes become one RPC. Two valid 600 KiB messages can produce an invalid >1 MiB frame. Split batches by serialized byte size or preserve lightweight flush boundaries. [GossipRpcPartsQueue.kt#L119-L149]

2. **[P2] Polling can retain disconnected peers.** The pump’s asynchronous supplier may execute after `onDisconnected()` removes the queue. `getQueue()` then recreates an empty entry keyed by the dead `PeerHandler`, with no later cleanup. Poll via a non-creating lookup or close/tombstone the pump. [AbstractRouter.kt#L334-L337]

Compatibility note: `RpcPartsQueue` is public; removing `takeMerged()` and adding abstract methods is source/binary breaking unless it is explicitly unsupported API.

Nashatyrev:
> 1. [P1] Unbounded batching can exceed `maxGossipMessageSize`.

Yeah, I have already hit this issue during my simulations and was going to address it on the next stage where we would need to reliably estimate the size of pending queue. So I'd suggest to postpone it till next PR as soon as it's not a regression 

> 2. [P2] Polling can retain disconnected peers.

Fair point, fixed it in the latest commit

tbenr (approved):
LGTM

PR #507, opened 2026-07-30
Gossipsub Backpressure: Size-Aware RPC Part Batching

Nashatyrev:
Follow-up to the backpressure queueing work: #505, #506

This PR makes pending RPC part queues aware of their estimated serialized size. Each queued RPC part now contributes a conservative standalone serialized-size estimate, and gossip batching uses that estimate together with the existing message-count limits when selecting parts for the next outbound RPC.

The estimate is likely larger than the final merged RPC because multiple parts can share protobuf wrapper messages once batched, but that makes it safe for pre-build splitting. The estimate may also be too conservative due to IHAVE topics, which are grouped in the final RPC. 

Changes include:
- Added queue-level serialized-size estimation for pending RPC parts.
- Applied maxGossipMessageSize when taking gossip RPC batches.
- Reset and update size accounting as parts are drained or aborted.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRpcPartsQueue.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRpcPartsQueue.kt
@@ -41,7 +41,7 @@ open class DefaultGossipRpcPartsQueue(
     private val params: GossipParams
 ) : DefaultRpcPartsQueue(), GossipRpcPartsQueue {
 
-    protected data class IHavePart(val messageId: MessageId, val topic: Topic) : AbstractPart {
+    protected data class IHavePart(val messageId: MessageId, val topic: Topic) : AbstractPart() {
         override fun appendToBuilder(builder: Rpc.RPC.Builder) {
             val ctrlBuilder = builder.controlBuilder
             val iHaveBuilder = ctrlBuilder.ihaveBuilderList
@@ -52,7 +52,7 @@ open class DefaultGossipRpcPartsQueue(
         }
     }
 
-    protected data class IWantPart(val messageId: MessageId) : AbstractPart {
+    protected data class IWantPart(val messageId: MessageId) : AbstractPart() {
         override fun appendToBuilder(builder: Rpc.RPC.Builder) {
             val ctrlBuilder = builder.controlBuilder
             val iWantBuilder = if (ctrlBuilder.iwantBuilderList.isEmpty()) {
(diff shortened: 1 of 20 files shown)


Nashatyrev:
P.S. this is the best balance I have found between simplicity/reliability and estimation precision. Anything more precise I tried was either slower or significantly more complicated and looked more error prone.

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRpcPartsQueue.kt
```diff
@@ -139,11 +144,17 @@ open class DefaultGossipRpcPartsQueue(
                 is GraftPart -> graftCount--
                 is PrunePart -> pruneCount--
             }
+            sizeLeft -= part.estimatedMaxSerializedSize
+            if (sizeLeft < 0) {
```
lucassaldanha, line 159: Is it possible to have a single part larger than `maxGossipMessageSize`? If so, we would break here with `partIdx = 0` and return null, but leave the part in the queue, that will end up "stuck" and the `publish()` future won't ever complete.

lucassaldanha: Maybe something like this:

```
if (sizeLeft < 0) {
      if (partIdx == 0) {
          // A part larger than maxGossipMessageSize would never fit, so emit it alone instead of returning null forever and stalling the peer's queue
          partIdx++
      }
      break
  }
```

Nashatyrev: Good catch! Thank you! 
However I would better propose to handle it earlier and just throw when an oversized message is submitted to the queue. Check the latest commit

lucassaldanha:
One thing that we might want to look at is at what point in `publish()` flow we are checking this message validity by size.

It looks like the message can end up in our seenMessages and mCache before it throws on `submitPublishmessage()`?

That would lead to:
- mCache advertising a message we can't send
- seenMessages poisoned, then throwing `MessageAlreadySeenException` instead of `TooLargeMessageException` for subsequent messages
- Our node starts sending IDONTWANT for a message that was never published

Maybe we should check for the size at the top of `broadcastOutbound() or even in `publish()` but earlier?

Nashatyrev:
> One thing that we might want to look at is at what point in `publish()` flow we are checking this message validity by size.

Good catch as well. 
This looks like a corner case while we already have similar case which occurs more often when `NoPeersForOutboundMessageException` is thrown. I would suggest to address this issue outside of this PR then: #510

PR #508, opened 2026-07-31
close UDP channel immediately on cancellation (QUIC)

tbenr:
- cancelled QUIC dials retained UDP sockets for 30,022 ms, not indefinitely.
- Cause: cancellation waited for the QUIC future instead of closing the parent UDP channel.
- Fix: QuicTransport.kt now closes that channel immediately.
- Added a blackhole regression test requiring release within 5 seconds.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
@@ -301,9 +301,11 @@ class QuicTransport(
         // (which runs a QUIC *server* codec), silently breaking the handshake. NAT-consistent
         // dialing for hole punching is handled separately by dialAsListener, which reuses the
         // listener socket directly.
-        val quicConnFuture: CompletableFuture<QuicChannel> = client.clone()
+        val udpBindFuture = client.clone()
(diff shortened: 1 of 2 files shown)


Nashatyrev (approved):
LGTM
Just a note for the future: there is basically no need to instantiate a UDP socket for every outbound QUIC connection, they could share a single UDP socket

tbenr:
> LGTM Just a note for the future: there is basically no need to instantiate a UDP socket for every outbound QUIC connection, they could share a single UDP socket

yes, true. Are there OS-level reasons it worth doing so? (thinking about dedicated buffers per socket)

"One implementation complication here is peer authentication: the current client codec and trust manager are created for a specific dial target. Sharing the parent socket would require preserving per-connection expected-peer validation rather than moving that trust decision to shared codec state."

Nashatyrev:
> Are there OS-level reasons it worth doing so?

I don't know to be honest. Basically UDP socket shouldn't be session aware, but there are qdisc which are aware. So it could be either way.

tbenr:
so ye, decoupling socket and authentication could enable different configurations (single socket for low resource consumption, pooled sockets, 1-1 like today)

PR #511, opened 2026-08-04
Fix/510 outbound publish admission

### LINKED ISSUE #510: Avoid retaining locally published messages after outbound publication fails
## Avoid retaining locally published messages after outbound publication fails

A locally initiated `publish()` currently adds the message to `seenMessages` before outbound publication is known to be admissible or possible. As a result, messages from failed publishes cannot be retried: the next call returns `MessageAlreadySeenException`.

This affects at least two cases:

1. **Oversized GossipSub messages**

   A publish can fail with `TooLargeMessageException` only when the message is added to the outbound RPC queue. At that point it has already been inserted into `seenMessages`; GossipSub may also have updated `mCache`, `lastPublished`, and queued IDONTWANT messages.

2. **No eligible outbound peers**

   When no peer is eligible for the message topics, `broadcastOutbound()` returns `NoPeersForOutboundMessageException`. The message has nevertheless already been marked seen (and added to `mCache`), so retrying after peers become available fails with `MessageAlreadySeenException`.

### Expected behavior

- Validate locally originated messages for outbound admission before mutating `seenMessages`, `mCache`, `lastPublished`, or outbound-control queues.
- An oversized message should consistently fail with `TooLargeMessageException` on every attempt.
- A message that fails solely because no suitable peers exist should be retryable once peers become available.
- Retain queue-level size validation as a safeguard for forwarded messages and non-publish RPC parts.

## Tests 

Below are the tests reproducing this behavior (in `GossipV1_1Tests`): 

```kotlin
    @Test
    fun `publishing too large message poisons seen cache`() {
        val msg = newMessage("topic1", 0L, "too-large".toByteArray())
        val standalonePublishSize = Rpc.RPC.newBuilder()
            .addPublish(msg.protobufMessage)
            .buildPartial()
            .serializedSize
        val test = TwoRoutersTest(GossipParams(maxGossipMessageSize = standalonePublishSize - 1))

        test.mockRouter.subscribe("topic1")

        assertThrows(TooLargeMessageException::class.java) {
            test.gossipRouter.publish(msg).getX()
        }

        assertThrows(MessageAlreadySeenException::class.java) {
            test.gossipRouter.publish(msg).getX()
        }
    }

    @Test
    fun `publishing without outbound peers poisons seen cache`() {
        val test = TwoRoutersTest()
        val msg = newMessage("topic1", 0L, "no-peers".toByteArray())

        assertThrows(NoPeersForOutboundMessageException::class.java) {
            test.gossipRouter.publish(msg).getX()
        }

        assertThrows(MessageAlreadySeenException::class.java) {
            test.gossipRouter.publish(msg).getX()
        }
    }
```

lucassaldanha:
Fixes #510.

`publish()` marked a message seen before outbound publication was known to be possible, so a failed publish could never be retried — the next attempt got `MessageAlreadySeenException`. For gossip, `mCache`, `lastPublished`, IDONTWANT and some peer queues had been mutated too, since the size limit was only enforced part-way through the per-peer submit loop.

Outbound publication is now two-phase:

1. **Admission** — message size is checked against `maxMsgSize` (`maxGossipMessageSize` for gossip) using the same estimate the outbound queue enforces, then `prepareOutboundPublish()` selects peers and throws if none is eligible. No router state is touched on either failing path.
2. **Execute** — only after the message is marked seen: `lastPublished`, `mCache`, IDONTWANT, per-peer queues, flush.

Result: an oversized message fails with `TooLargeMessageException` on every attempt, and a publish that failed only for lack of peers is publishable again once peers appear. The queue-level size guard stays as the safeguard for forwarded messages and control parts; inbound handling is unchanged.

### Breaking (advanced API)

`AbstractRouter.broadcastOutbound(msg): CompletableFuture<Unit>` is replaced by `AbstractRouter.prepareOutboundPublish(msg): OutboundPublish`. Users of `GossipRouterBuilder`, `GossipRouter` or `FloodRouter` are unaffected; direct `AbstractRouter` subclasses need:

```kotlin
override fun prepareOutboundPublish(msg: PubsubMessage) = OutboundPublish {
    /* previous broadcastOutbound body */
}
```

Throw from prepareOutboundPublish for conditions that make the message unpublishable at all.

### Other observable changes

- A publish that is both oversized and has no eligible peers now fails with TooLargeMessageException rather than NoPeersForOutboundMessageException.
- An oversized message whose peers had all sent IDONTWANT previously succeeded silently; it now throws.
- A gossip publish with no eligible peers no longer adds the message to mCache or updates lastPublished.
- Retryability covers admission failures only: a publish that fails after its parts were queued (peer disconnect, all writes failing) still leaves the message seen. Documented in the KDoc.

### Tests

GossipV1_1Tests: oversized publish fails identically on retry and leaves no mCache entry; a message of exactly maxGossipMessageSize still publishes; a no-peers publish succeeds after the peer subscribes.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -71,9 +71,28 @@ abstract class AbstractRouter(
                 completedExceptionally(MessageAlreadySeenException("Msg: $msg"))
             } else {
                 messageValidator.validate(msg) // check ourselves not to be a bad peer
+                validateOutboundMessageSize(msg)
+                val publication = prepareOutboundPublish(msg)
                 seenMessages[msg] = Optional.of(ValidationResult.Valid)
-                broadcastOutbound(msg)
+                publication.publish()
             }
+        }
+    }
+
+    /**
+     * Checks that a locally originated message can be sent as a single outbound RPC.
+     *
+     * Called before any router state is mutated, so an oversized message is never retained: every
+     * publish attempt fails with [TooLargeMessageException] instead of the second and later attempts
+     * failing with [MessageAlreadySeenException].
+     */
+    protected open fun validateOutboundMessageSize(msg: PubsubMessage) {
+        val estimatedSize = estimateStandalonePublishRpcSize(msg.protobufMessage)
+        if (estimatedSize > maxMsgSize) {
+            throw TooLargeMessageException(
+                "Locally published message estimated serialized size $estimatedSize " +
+                    "exceeds max message size $maxMsgSize: $msg"
+            )
         }
     }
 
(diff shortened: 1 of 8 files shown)


Nashatyrev:
I'm leaning to complete the backpressure changes first and then get back to this issue

PR #512, opened 2026-08-04
Gossip backpressure: Prioritize Gossip RPC parts and route IDONTWANT through backpressure queue

### LINKED ISSUE #440: Investigate impact of splitting control extension messages from extensions messages
We are sending control extensions messages using the current implementation of `GossipRpcPartsQueue`. We need to investigate the scenario where we have on the queue both a control extension message enabling a particular extension X, and a X extension message.

In the case of splitting, if the X extension message ends up being sent first, we might have an issue where the peer won't know we support that extension because the control extension message with the flag for X support will be sent later.

Questions:
- Can this happen in practice?

Nashatyrev:
Followup for #505, #506, #507

This PR extends Gossip outbound RPC batching to support priority-based part queues.

Changes include:

- Route outbound `IDONTWANT` through `GossipRpcPartsQueue` instead of direct `send(...)`, so it follows the same backpressure-aware path as other RPC parts.
- Add `IDONTWANT` part support and enforce `maxIDontWantMessageIds` in outbound batch splitting.
- Split Gossip RPC parts into priority buckets:
    - highest priority: `IDONTWANT` and control extensions
    - state/control: subscriptions, `GRAFT`, `PRUNE`
    - bulk: publish, `IHAVE`, `IWANT`
- Ensure `takeBatch()` only drains the highest non-empty priority bucket and does not mix priorities in one RPC.
- Extract `AbstractRpcPartsQueue` so default and Gossip queues can share size accounting and promise merging while using different storage layouts.

Resolves #440.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/RpcPartsQueue.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/RpcPartsQueue.kt
@@ -1,7 +1,6 @@
 package io.libp2p.pubsub
 
 import io.libp2p.etc.types.forward
-import io.libp2p.pubsub.DefaultRpcPartsQueue.AbstractPart
 import pubsub.pb.Rpc
 import java.util.concurrent.CompletableFuture
 
@@ -79,12 +78,7 @@ interface RpcPartsQueue {
     fun estimateMaxSerializedSize(): Int
 }
 
-/**
- * Default [RpcPartsQueue] implementation
- *
- * NOT thread safe
- */
-open class DefaultRpcPartsQueue : RpcPartsQueue {
+abstract class AbstractRpcPartsQueue : RpcPartsQueue {
 
     protected abstract class AbstractPart {
 
@@ -128,16 +122,23 @@ open class DefaultRpcPartsQueue : RpcPartsQueue {
         }
     }
 
-    protected open val parts = mutableListOf<AbstractPart>()
     private var estimatedMaxSerializedSizeAccum: Int = 0
 
-    protected open fun addPart(part: AbstractPart) {
-        parts += part
+    protected abstract fun addPart(part: AbstractPart)
+
+    protected open fun addPartSize(part: AbstractPart) {
         estimatedMaxSerializedSizeAccum += part.estimatedMaxSerializedSize
     }
-    protected fun onPartsRemoving(removedParts: List<AbstractPart>) {
-        estimatedMaxSerializedSizeAccum -= removedParts.sumOf { it.estimatedMaxSerializedSize }
+    protected fun removePartsSize(removedParts: List<AbstractPart>) {
+        removedParts.forEach {
+            removePartSize(it)
+        }
     }
+    protected fun removePartSize(removedPart: AbstractPart) {
+        estimatedMaxSerializedSizeAccum -= removedPart.estimatedMaxSerializedSize
+    }
+
+    override fun estimateMaxSerializedSize(): Int = estimatedMaxSerializedSizeAccum
 
     override fun addPublish(message: Rpc.Message) {
         addPart(PublishPart(message))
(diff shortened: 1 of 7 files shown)


tbenr:
**Findings**
1. **P1: zero `IDONTWANT` limit stalls all outbound traffic.** `GossipRpcPartsQueue.kt:193` globally requires `iDontWantCount > 0`, even for publish-only batches. With the accepted value `maxIDontWantMessageIds = 0`, `takeBatch()` returns `null` while non-empty, leaving publish promises unresolved.

```kotlin
    @Test
    fun queuedIDontWantIsFlushedWhenAllPublishPeersOptedOut() {
        val test = startSingleTopicNetwork(
            params = GossipParams(iDontWantMinMessageSizeThreshold = 5),
            mockRouterCount = 2
        )
        val msg = newMessage("topic1", 0L, "Hello".toByteArray())

        test.mockRouters.forEach { router ->
            router.sendToSingle(
                Rpc.RPC.newBuilder().setControl(
                    Rpc.ControlMessage.newBuilder().addIdontwant(
                        Rpc.ControlIDontWant.newBuilder().addMessageIDs(msg.messageId.toProtobuf())
                    )
                ).build()
            )
        }
        test.fuzz.timeController.addTime(100.millis)
        val inboundCounts = test.mockRouters.map { it.inboundMessages.size }

        test.gossipRouter.publish(msg)
        test.fuzz.timeController.addTime(100.millis)

        test.mockRouters.zip(inboundCounts).forEach { (router, count) ->
            val newIDontWants = router.inboundMessages.drop(count)
                .flatMap { it.control.idontwantList }
                .flatMap { it.messageIDsList }
                .map { it.toWBytes() }
            assertThat(newIDontWants).contains(msg.messageId)
        }
    }
 ```

seems like the following is intentional?

2. **P2: queued `IDONTWANT` can miss its pump notification.** `GossipRouter.kt:543` queues messages, but when every publish target previously opted out, the empty branch returns without `flushAllPending()`. Delivery waits until another flush, usually the next heartbeat. A regression test failed on the PR and passed after moving the flush before the branch.

```kotlin
    @Test
    fun `zero IDONTWANT limit does not block unrelated publish`() {
        val partsQueue = TestGossipQueue(
            GossipParamsBuilder()
                .maxIDontWantMessageIds(0)
                .maxIHaveLength(Int.MAX_VALUE)
                .build()
        )

        partsQueue.addPublish(createRpcMessage("topic", "data"))

        assertThat(partsQueue.takeBatch()).isNotNull()
    }
```

image



we dont care about this:
3. **P2: queue extension APIs become source/binary incompatible.** `GossipRpcPartsQueue.kt:20` adds an abstract interface method, while `GossipRpcPartsQueue.kt:49` changes the superclass and protected API. Existing custom implementations may throw `AbstractMethodError`; subclasses using `parts` or `DefaultRpcPartsQueue.AbstractPart` break. This needs compatibility shims or an explicitly versioned breaking release.

Nashatyrev:
> 1. **P1: zero `IDONTWANT` limit stalls all outbound traffic.** `GossipRpcPartsQueue.kt:193` globally requires `iDontWantCount > 0`, even for publish-only batches. With the accepted value `maxIDontWantMessageIds = 0`, `takeBatch()` returns `null` while non-empty, leaving publish promises unresolved.

I would prefer to not allow `maxIDontWantMessageIds = 0` because it doesn't make much sense here as it's just DoS protection measure. Addressed by 1ab4ec8

> 2. **P2: queued `IDONTWANT` can miss its pump notification.** `GossipRouter.kt:543` queues messages, but when every publish target previously opted out, the empty branch returns without `flushAllPending()`. Delivery waits until another flush, usually the next heartbeat. A regression test failed on the PR and passed after moving the flush before the branch.

Addressed with 83bd2d7

> 3. P2: queue extension APIs become source/binary incompatible.

We are not API backward compatible here

lucassaldanha (approved):

PR #513, opened 2026-08-04
Gossip backpressure: detect and handle slow peers under outbound backpressure

Nashatyrev:
Followup for #505, #506, #507, #512

## Summary

- Detect peers whose outbound RPC queues remain above a configurable byte threshold for consecutive heartbeats.
- First drop low-priority bulk data while preserving higher-priority control messages.
- Drop all remaining queued data if removing low-priority data does not reduce the queue below the threshold.
- Fail affected publish futures with `DroppedRpcPartsException`.
- Notify router event listeners and apply a behavior penalty through `DefaultGossipScore`.
- Keep slow-peer connections open after trimming their queues.
- Expose slow-peer thresholds through `GossipParams` and `GossipParamsBuilder`; detection remains effectively disabled by default.

## Configuration note

The pending-bytes threshold should not be set too high, because that would allow excessive outbound data to accumulate before backpressure handling starts. It is expected that the threshold may be exceeded occasionally under normal bursty traffic; the consecutive-heartbeat threshold prevents these temporary bursts from classifying a peer as slow.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -60,6 +60,7 @@ abstract class AbstractRouter(
             .map { it.key }
         fun getOrCreateQueue(peer: PeerHandler) = map.computeIfAbsent(peer) { queueFactory() }
         fun getExistingQueue(peer: PeerHandler) = map[peer]
+        fun getQueues(): Map<PeerHandler, TPartsQueue> = map
         fun onDisconnected(peer: PeerHandler) {
             map.remove(peer)?.abort(ConnectionClosedException())
         }
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt
@@ -62,7 +62,7 @@ abstract class AbstractRouter(
         fun getExistingQueue(peer: PeerHandler) = map[peer]
         fun getQueues(): Map<PeerHandler, TPartsQueue> = map
         fun onDisconnected(peer: PeerHandler) {
-            map.remove(peer)?.abort(ConnectionClosedException())
+            map.remove(peer)?.dropAll(ConnectionClosedException())
         }
     }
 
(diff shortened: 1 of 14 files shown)


tbenr:
[P1] Pressure tracking fails above 256 slow peers. The bounded map churns counters each heartbeat, so with 257 pressured peers none reaches threshold. Use state bounded by connected queues instead. [GossipRouter.kt:127](https://github.com/libp2p/jvm-libp2p/blob/8b82529755ed543f42a03e0aacec61a78255bd54/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt#L127)

the patch contains the regression test proving the issue and a possible solution
[pr-513-queue-state.patch](https://github.com/user-attachments/files/30739392/pr-513-queue-state.patch)

----

Use a smaller equivalent: capacity **3**, but **4** pressured peers (`A B C D`), with threshold `2`.

**Heartbeat 1**

```text
A:1, B:1, C:1, D:1
```

Because capacity is 3, adding `D` evicts `A`.

Final map:

```text
B:1, C:1, D:1
```

**Heartbeat 2**

Queues are visited again in stable order:

```text
A added -> evicts B
B added -> evicts C
C added -> evicts D
D added -> evicts A
```

Final map is again:

```text
B:1, C:1, D:1
```

Every counter is recreated at `0` and incremented only to `1`; none reaches `2`.

PR has the same pattern with capacity `256` and `257` pressured queues. This only affects `slowPeerHeartbeatThreshold > 1`; threshold `1` triggers immediately.

tbenr:
another approach could be avoid the limited LRU and just accept the limit from the max peers

private val slowPeerQueuePressure =
    mutableMapOf<PeerHandler, SlowPeerQueuePressure>()
    
    
but yeah, in general this is an edge case\an adversarial condition

Nashatyrev:
https://github.com/libp2p/jvm-libp2p/blob/8b82529755ed543f42a03e0aacec61a78255bd54/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipRouter.kt#L53-L58

Oh yeah, these constants suck. This 256 limit is effectively max number of peers which doesn't look very conservative to be hardcoded. I can easily imagine beefy nodes with 1000 and more peers.

Nashatyrev:
> another approach could be avoid the limited LRU and just accept the limit from the max peers

The problem is that we have no any limit on the number of peers in gossip. 
LRU limits are just kind of double defence against accidental bug which causes unbounded map growth

Nashatyrev:
I would suggest to address this in a separate PR. Here is the issue on this: https://github.com/libp2p/jvm-libp2p/issues/514

tbenr (approved):
LGTM

PR #521, opened 2026-08-26
Bound control-plane size of inbound and outbound pubsub RPCs

lucassaldanha:
**Summary**

Our pre-decode validator caps the number of messageIDs inside ihave/iwant, but nothing caps the number of envelopes. Envelopes are cheap on the wire and not cheap to decode, so the counters we have don't bound what a frame actually costs us to materialise.

Similar approaches by other libp2p implementations:
- https://github.com/libp2p/go-libp2p-pubsub/pull/707
- https://github.com/libp2p/rust-libp2p/pull/6468

**What's in here:**

- new PubsubRpcLimits.maxControlMessageSize, appended to the end of the constructor so positional callers still compile
- new GossipParams.maxControlMessageSize, default 256 KiB, plumbed into GossipRouter.rpcLimits
- validator accumulates control-plane bytes during the walk it already does, no second pass, and rejects before ProtobufDecoder runs
- charging rule: subscriptions and control in full, and so is every other top-level field, known or unknown. publish and RPC.partial are charged for their envelope with only their opaque payload lengths exempt (Message.data, partialMessage, partsMetadata)
- unknown fields are charged, never rejected. go retains and re-emits them, so rejecting would break interop the next time anyone adds a field
- takeBatch now bounds control bytes on the way out too. PublishPart, IHavePart and IWantPart share a priority list and get merged into one RPC, so without this we could emit frames we would reject ourselves
- new GossipParams.maxIDontWantMessageIdsPerRpc, default 5000, so a default configured node doesn't put 50k ids in a single RPC. Teku already sets maxIDontWantMessageIds to 5000 so it is unaffected

**Note**

We chose 256KiB by measuring our own batcher emissions through different testing scenarios. Upon further testing we can revisit this value.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/builders/GossipParamsBuilder.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/builders/GossipParamsBuilder.kt
@@ -82,6 +82,8 @@ class GossipParamsBuilder {
 
     private var slowPeerHeartbeatThreshold: Int? = null
 
+    private var maxControlMessageSize: Int? = null
+
     init {
         val source = GossipParams()
         this.D = source.D
@@ -115,6 +117,7 @@ class GossipParamsBuilder {
         this.iDontWantTTL = source.iDontWantTTL
         this.slowPeerPendingBytesThreshold = source.slowPeerPendingBytesThreshold
         this.slowPeerHeartbeatThreshold = source.slowPeerHeartbeatThreshold
+        this.maxControlMessageSize = source.maxControlMessageSize
     }
 
     fun D(value: Int): GossipParamsBuilder = apply { D = value }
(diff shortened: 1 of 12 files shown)


libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/builders/GossipParamsBuilder.kt
```diff
@@ -82,6 +82,10 @@ class GossipParamsBuilder {
 
     private var slowPeerHeartbeatThreshold: Int? = null
 
+    private var maxControlMessageSize: Int? = null
+
+    private var maxIDontWantMessageIdsPerRpc: Int? = null
```
Nashatyrev, line 87: nit: do we really need one more cap parameter? Why can't we have the same per-PRC limit as `maxIDontWantMessageIds`?

Nashatyrev (approved):
Look good to me. Just one nit

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/builders/GossipParamsBuilder.kt
lucassaldanha: they are actually used at different places, but in the following PR, I'm gonna clean a bunch of this usage and it will be nicer. Just trying to avoid changing the API on this change.

PR #524, opened 2026-08-31
fix(quic): run delegated TLS tasks inline, not on shared workerGroup

### LINKED ISSUE #523: Teku crashes unexpectedly while handling QUIC connections
Node crashed with `SIGSEGV` inside `io.netty.handler.codec.quic.BoringSSL.SSL_getTask(long)`

## Environment

- Teku `26.8.0`
- OpenJDK `25.0.4`
- Netty QUIC `4.2.16.Final`
- jvm-libp2p `1.3.5`
- QUIC enabled
- Process uptime before the crash: approximately 35 hours

## Findings

This was not caused by an OOM, GC pressure, host restart, or execution-client failure.

Core-dump analysis showed:

- The crashing event-loop thread was executing `BoringSSL.SSL_getTask`.
- Another event-loop thread was concurrently executing `QuicheQuicConnection.sslTask()` on the same connection object, following `quiche_conn_send`.
- The native `SSL_getTask` call returned an invalid-looking JNI handle (`RAX=0x5cdf14b505d41ada`), which HotSpot crashed while dereferencing.
- The native implementation returns `ssl_task->task`, suggesting that the task structure or its JNI reference may have been replaced or freed concurrently.

jvm-libp2p configures the QUIC client and server with:

```java
.sslTaskExecutor(workerGroup)
```

This shared multithreaded executor appears to allow delegated TLS task retrieval to overlap with QUIC send/receive processing or connection teardown. Netty accepts a generic `Executor` and does not document a same-thread requirement.

The evidence therefore points to a possible race/use-after-free involving delegated TLS task state. The exact free operation has not been captured, so this is strongly supported but not yet conclusively proven.

## Reproducibility

The crash is timing-dependent and is not yet deterministically reproducible. It may occur during concurrent QUIC handshakes combined with:

- Connection cancellation or timeout
- Duplicate connection handling
- Connection teardown
- High peer or handshake churn

The observed node ran for approximately 35 hours before the crash, and there was no advance warning.

## Impact

The race causes complete JVM termination and validator-node downtime.
A possible short-term library workaround is to remove `.sslTaskExecutor(workerGroup)` and execute delegated TLS tasks inline or on the connection's owning event loop.

Netty `4.2.17` appears to retain the same native implementation, so upgrading alone does not appear to address the issue.

## Long-Term Solution

The permanent solution should primarily be implemented in Netty QUIC:

- Serialize access to delegated TLS tasks, QUIC send/receive operations, and connection teardown.
- Keep JNI task references alive until execution has completed.
- Prevent native connection or SSL state from being destroyed while delegated tasks remain pending.
- Make delegated task handling safe regardless of the executor supplied by callers.

jvm-libp2p should also avoid dispatching connection-specific TLS tasks directly to a shared multithreaded `workerGroup`. Tasks should run inline or be serialized through the connection's owning event loop until Netty provides stronger lifecycle guarantees.

lucassaldanha:
## Summary

Fixes #523 — a use-after-free race that crashed a mainnet Teku node with `SIGSEGV` in `BoringSSL.SSL_getTask`.

- `QuicheQuicConnection.sslTask()`/`free()` (in netty's QUIC codec) synchronize on the connection object only around *fetching* the delegated task / freeing native state — not around the task's actual native execution.
- `QuicheQuicChannel.connectionSend()` dispatches delegated tasks to whatever thread `sslTaskExecutor` picks, unless that executor is `null`/`ImmediateExecutor`/`ImmediateEventExecutor`, in which case tasks run inline on the calling event-loop thread.
- `QuicTransport` passed the shared multi-thread `workerGroup` as `sslTaskExecutor` for both the client-dial and server codecs, so a delegated task could keep executing on a background worker thread while the connection's own event-loop thread concurrently called `free()` (`SSL_cleanup`/`quiche_conn_free`) — a use-after-free on the native `ssl`/`connection` pointers.

Switches both codecs to `ImmediateEventExecutor.INSTANCE`, so delegated tasks always run inline on the connection's own event-loop thread — the same workaround suggested in #523. jvm-libp2p only configures Ed25519/ECDSA keys with a synchronous JDK trust manager (no async private-key ops, no OCSP), so there is no slow delegated task that would risk blocking the event loop.

Reference from Netty's code: https://github.com/netty/netty/blob/netty-4.2.16.Final/codec-classes-quic/src/main/java/io/netty/handler/codec/quic/QuicheQuicChannel.java#L1194-L1196

## Test plan

- [x] New `QuicSslTaskExecutorTest`: drives a real client↔server QUIC handshake, then reflects into the live `QuicheQuicChannel.sslTaskExecutor` field on both sides and asserts it's `ImmediateEventExecutor.INSTANCE`. The underlying native race is timing-dependent and not deterministically reproducible in a unit test, so this asserts the fix's actual contract (no cross-thread task dispatch) instead.
- [x] `./gradlew :libp2p:test --tests "io.libp2p.transport.quic.*"` — all QUIC tests pass.
- [x] `./gradlew :libp2p:spotlessCheck` — clean.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/transport/quic/QuicTransport.kt
@@ -31,6 +31,7 @@ import io.netty.channel.nio.NioIoHandler
 import io.netty.channel.socket.nio.NioDatagramChannel
 import io.netty.handler.codec.quic.*
 import io.netty.handler.ssl.ClientAuth
+import io.netty.util.concurrent.ImmediateEventExecutor
 import org.slf4j.LoggerFactory
 import java.net.Inet6Address
 import java.net.InetSocketAddress
@@ -281,7 +282,14 @@ class QuicTransport(
         val sslContext = quicSslContext(true, trustManager)
         val requestsHandler = QuicClientCodecBuilder()
             .sslEngineProvider { q -> sslContext.newEngine(q.alloc()) }
-            .sslTaskExecutor(workerGroup)
+            // Delegated TLS tasks must run on the connection's own event-loop thread, not a
+            // shared multi-thread pool: Netty QUIC only serializes task retrieval (SSL_getTask)
+            // against connection teardown (SSL_cleanup/quiche_conn_free), not the task's native
+            // execution itself, so a task run on a background thread can race a concurrent close
+            // and use freed native SSL/quiche state (see
+            // https://github.com/libp2p/jvm-libp2p/issues/523). ImmediateEventExecutor
+            // makes Netty run tasks inline on the calling (event-loop) thread instead.
+            .sslTaskExecutor(ImmediateEventExecutor.INSTANCE)
             .maxIdleTimeout(config.idleTimeout.toMillis(), TimeUnit.MILLISECONDS)
             .initialMaxData(config.maxConnectionData)
             .initialMaxStreamsBidirectional(config.maxStreamsBidirectional)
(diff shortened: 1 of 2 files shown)


tbenr (approved):
LGTM

PR #525, opened 2026-09-01
Consolidate per type counters on gossip

lucassaldanha:
Consolidates the per-type repeated-field counters into a control-byte budget. The old per-type counters existed to stop a peer forcing us to allocate too many objects. The new byte budget already stops that, for every field, by size instead of by count.

`maxIHaveLength`, `maxIHaveMessages`, `maxPeersAcceptedInPruneMsg`, `maxIDontWantMessageIds`, `maxIDontWantMessageIdsPerRpc`, and `maxControlMessageSize` all stay untouched; they are still being used outside the decode/validation/batch-split path (IWANT amplification tracking, the heartbeat IDONTWANT cache, PX peer-list truncation).

**Breaking change**: `GossipParams` and `GossipParamsBuilder` lose `maxSubscriptions`, `maxIWantMessageIds`, `maxGraftMessages` and `maxPruneMessages`. `PubsubRpcLimits` loses those four plus `maxIHaveMessageIds`, `maxPeersPerPruneMessage`, `maxIDontWantMessages`, `maxIDontWantMessageIds` and `rejectEmptyIDontWantEntries`. Teku sets several of them explicitly in `LibP2PParamsFactory` - that's a separate follow-up PR once this lands, not included here.

`maxSubscriptionsPerRpc` is added to cover the outbound half of `maxSubscriptions`: peers enforce subscriptions by count and typically drop the whole RPC when it is exceeded, so a router announcing many topics at once still needs a count-based split.

Tracked on https://github.com/Consensys/teku-internal/issues/317


tbenr:
**Findings**
1. **[P2] Preserve the dedicated IWANT cap.** [GossipRouter.kt:308]
   `handleIWant` queues one publish per matching ID, so this limit also controls response amplification, not just allocations. The 256 KiB budget permits roughly 11.9k 20-byte IDs versus Teku’s former 5k cap, and more with shorter custom IDs. Retain a processing or per-peer response limit.

2. **[P3] Update the PRUNE peer-limit contract.** [GossipParams.kt:214]
   Oversized PRUNE peer lists are no longer rejected; the router processes the PRUNE and only truncates PX callbacks. The public KDoc still says this is the maximum “allowed” in an incoming message.

libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/GossipParams.kt
```diff
@@ -241,11 +225,6 @@ data class GossipParams(
      */
     val pruneBackoff: Duration = 1.minutes,
 
-    /**
-     * [maxPruneMessages] is the maximum number of prune messages allowed per gossip message
-     */
-    val maxPruneMessages: Int? = null,
-
```
Nashatyrev, line 248: For all these removed caps: don't other clients have them? If yes, then we may start producing invalid messages with too many subscriptions/IWANT/GRAFT/PRUNE messages

lucassaldanha: That's a good point. From what I could see only the subscription count is something that we might need to keep, to stay in line with with others. Prysm is using a 500 max subscription limit. 

Even in TEku we have MaxCountTopicSubscriptionFilter checking if it is more than 200. We are being conservative, assuming one supernode would need 128 subscriptions for all columns subnets + 64 attestation subnets + other global topics. I am not wondering if we should bump it to match go's 500 limit.

For others like GRAFT/PRUNE/IWANT, both go-libp2p and rust-libp2p have used a similar byte budget approach so I believe we are fine.

lucassaldanha:
> 1. **[P2] Preserve the dedicated IWANT cap.** [GossipRouter.kt:308]
>    `handleIWant` queues one publish per matching ID, so this limit also controls response amplification, not just allocations. The 256 KiB budget permits roughly 11.9k 20-byte IDs versus Teku’s former 5k cap, and more with shorter custom IDs. Retain a processing or per-peer response limit.

I'm not sure if I agree with this assessment. For a per-RPC protection, both the count cap and the byte budget offer the same type of protection; the "total" number of accepted requests might not be exactly the same, but the total allocated memory is bounded, so there is an upper limit. Happy to be proven wrong with a test case or something that shows that they don't.

However, there is a gap on the current implementation, something like "amplification over time", when we receive multiple IWANT/IWANT from the same peer etc. This would be handled by https://github.com/Consensys/teku-internal/issues/324

> 2. **[P3] Update the PRUNE peer-limit contract.** [GossipParams.kt:214]
>    Oversized PRUNE peer lists are no longer rejected; the router processes the PRUNE and only truncates PX callbacks. The public KDoc still says this is the maximum “allowed” in an incoming message.

Will fix it :)

Nashatyrev (approved):
LGTM
But I would like to have another approval here

lucassaldanha:
> `handleIWant` queues one publish per matching ID, so this limit also controls response amplification, not just allocations. The 256 KiB budget permits roughly 11.9k 20-byte IDs versus Teku’s former 5k cap, and more with shorter custom IDs. Retain a processing or per-peer response limit.

Ok, after diving deeper into this, I think we have the following:
- Regardless of the previous cap on 5k IWANTs or the current byte-budget approach, we are still prone to the same issue, this is due to the fact that we aren't limiting the output of replies to IWANT messages.
- Even in our previous 5k limit, we would basically have the same issue if a peer submitted two requests with 5k each (verified with test), so we are not better by reinstating the prior limit. What the limit did was ensure a "per-RPC" cap, but not an "by-time" cap, so the problem would still be there.
- We have other mechanisms that bound the amount of messages we are sending back to a peer (`gossipRetransmission = 3` and the recently added slow peer handingl) that mitigate some of the issue.

To be clear the gap is real and documented https://github.com/Consensys-Incorporated/teku-internal/issues/324

Issue #526, opened 2026-09-03
Cross-impl gossipsub: direction-dependent silent stream drop in P2PServiceSemiDuplex (subscription/GRAFT lost vs go-libp2p)


lucassaldanha:
## Summary

Cross-implementation gossipsub interop testing (against go-libp2p-pubsub) shows two direction-dependent failures that both look like the same root cause in `P2PServiceSemiDuplex`: depending on which side dials, either subscription notifications or GRAFT messages silently fail to reach the peer, while the underlying connection stays up.

## Observed symptoms

gossipsub-smoke-tests (Ethereum gossipsub profile, go-libp2p-pubsub v0.13.1 as the peer).

1. **jvm as dialee never sees the dialer's subscription.** Peer (go) dials jvm, both subscribe to a topic; jvm never observes the peer's subscription announcement (times out after 7s, consistently reproducible).
2. **jvm as dialer never grafts into the dialee's mesh, from the dialee's perspective.** jvm dials out to a peer (go); the peer's mesh never comes to contain jvm's peer id (times out after 10.5s, consistently reproducible). The reverse (jvm as dialee grafting) works.

Both are 100% reproducible on every run, not flaky. jvm-jvm and go-go pairings pass cleanly; only jvm-go/go-jvm show these five test failures (subscription exchange, mesh graft in both directions, full-mesh publish in both directions - the last two are downstream of the same underlying misses).

## Suspected root cause

`P2PServiceSemiDuplex` models each connection as two independent unidirectional streams (an "initiator" one we open, an "acceptor" one the peer opens), merged into one `SDPeerHandler`:

https://github.com/libp2p/jvm-libp2p/blob/da462a6d5744c9e73cbb1b0ac642f07d731eefbf/libp2p/src/main/kotlin/io/libp2p/etc/util/P2PServiceSemiDuplex.kt#L34-L56

`Gossip.handleConnection` has both peers proactively open a gossipsub stream on every established connection, regardless of which side dialed at the transport layer - i.e. a simultaneous-open at the stream level on every peering:

https://github.com/libp2p/jvm-libp2p/blob/da462a6d5744c9e73cbb1b0ac642f07d731eefbf/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt#L63-L65

[...]

PR #528, opened 2026-09-07
fix(gossip): open a single outbound stream per peer

lucassaldanha:
Found while investigating #526. It does not fix that issue, which turned out to be a configuration bug in the reporter's test harness. This is a separate, real bug that the investigation turned up, reproduced independently against go-libp2p. See the last section.

libp2p allows more than one connection to the same peer (https://github.com/libp2p/specs/blob/master/pubsub/README.md#stream-management), most commonly when both sides dial each other at the same time. `Gossip.handleConnection` opened a gossipsub stream on every connection, so with two connections we open two outbound streams to the same peer. 

go-libp2p-pubsub keeps a single inbound stream per peer id and `Reset()`s the **OLDER** one when a duplicate arrives (`pubsub.go`, `handleNewStream`), `P2PServiceSemiDuplex.streamAdded` pairs streams by peer id and keeps the **OLDER** one, so it closes our second. Both of our outbound streams die, one killed by each side. `streamDisconnected` then tears the whole `PeerHandler` down, the peer's respawned writer creates a fresh inbound-only handler, and nothing ever opens a replacement outbound stream. The peering stays up but is permanently one-way: we receive the peer's SUBSCRIBE and GRAFT, ours never leave.

Reproduced against go-libp2p v0.49.0 and go-libp2p-pubsub v0.13.1 over TCP + Noise + Yamux. Single-connection peering is clean in every ordering I tried, either side dialing. With a synchronised simultaneous dial it failed in 1 of 3 runs, and note that neither connection closed, go reset the stream rather than the connection:

```
### RESULT simultaneous-connect-0: sub=true goMesh=false jvmMesh=true
### jvm connections=2
### jvm activePeers=[]
### jvm peerStreams=in=true out=false
```

### Changes

- `Gossip` now tracks the live connections of each peer and opens the gossipsub stream on one of them only, so we keep a single outbound stream per peer id the way go-libp2p-pubsub does and neither side ends up with a same-direction duplicate to drop.
- The stream is reopened on another live connection when the connection carrying it closes or the peer resets it, which is the equivalent of go's "peer declared dead but still connected; respawning writer" path in `handleDeadPeers`.
- The duplicate-stream and third-stream exceptions in `P2PServiceSemiDuplex` name the dropped stream and the streams the peer already had, because a one-way peering caused by a dropped stream is only diagnosable if the error says which stream went. Semantics are unchanged, and note these are logged at `debug` by `AbstractRouter.onServiceException`, so at production log levels this failure currently leaves no trace at all.
- `GossipMultipleConnectionsTest` covers the handover: with two connections to a peer, closing the one carrying our outbound stream has to leave the peering working. It fails on `develop`, where both routers drop to zero gossip peers.

### DIFF
--- a/libp2p/src/main/kotlin/io/libp2p/etc/util/P2PServiceSemiDuplex.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/etc/util/P2PServiceSemiDuplex.kt
@@ -41,11 +41,17 @@ abstract class P2PServiceSemiDuplex(executor: ScheduledExecutorService) : P2PSer
             when {
                 peerHandler.otherStreamHandler != null -> {
                     streamHandler.closeAbruptly()
-                    throw BadPeerException("Duplicate steam for peer ${peerHandler.peerId}. Closing it silently")
+                    throw BadPeerException(
+                        "Third stream $stream for peer ${peerHandler.peerId} which already has " +
+                            "${peerHandler.streamHandler.stream} and ${peerHandler.otherStreamHandler?.stream}. Closing it"
+                    )
                 }
                 peerHandler.streamHandler.stream.isInitiator == stream.isInitiator -> {
                     streamHandler.closeAbruptly()
-                    throw BadPeerException("Duplicate stream with initiator = ${stream.isInitiator} for peer ${peerHandler.peerId}")
+                    throw BadPeerException(
+                        "Duplicate stream $stream with initiator = ${stream.isInitiator} for peer " +
+                            "${peerHandler.peerId} which already has ${peerHandler.streamHandler.stream}. Closing it"
+                    )
                 }
                 else -> {
                     peerHandler.otherStreamHandler = streamHandler
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
@@ -24,16 +24,9 @@ class Gossip @JvmOverloads constructor(
 
     private val logger = LoggerFactory.getLogger(Gossip::class.java)
 
-    /**
-     * Live connections per peer.
-     *
-     * libp2p allows more than one connection to the same peer - most commonly when both sides dial
-     * each other at the same time - while gossipsub, like go-libp2p-pubsub, keeps a single outbound
-     * stream per peer id. Opening a stream on every connection makes both peers see same-direction
-     * duplicates and drop them, which leaves the peering permanently unable to write. So the stream
-     * is opened on one connection only and reopened on another one when that connection goes away.
-     */
-    private val peerConnections = mutableMapOf<PeerId, MutableList<Connection>>()
+    private val outboundStreams = SingleOutboundStreamPerPeer { conn ->
+        conn.muxerSession().createStream(listOf(this)).stream
+    }
 
     fun updateTopicScoreParams(scoreParams: Map<String, GossipTopicScoreParams>) {
         router.score.updateTopicParams(scoreParams)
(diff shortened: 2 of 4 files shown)


Nashatyrev:
Relates to https://github.com/libp2p/go-libp2p/issues/634 (Multiple connection pruning)

Nashatyrev:
nit: I would suggest extracting the essential functionality into a dedicated class to keep the Gossip class more manageable. Something like this: 

```diff
Index: libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt
--- a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt	(revision 12862199375cfb697d90c4bde8be9ee1f9f7cdd4)
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/Gossip.kt	(date 1789055062391)
@@ -24,6 +24,14 @@
 
     private val logger = LoggerFactory.getLogger(Gossip::class.java)
 
+    private val outboundStreams = SingleOutboundStreamPerPeer { conn ->
+         conn.muxerSession().createStream(listOf(this)).stream
+    }
+
     fun updateTopicScoreParams(scoreParams: Map<String, GossipTopicScoreParams>) {
         router.score.updateTopicParams(scoreParams)
     }
@@ -76,9 +84,7 @@
             }
         }
 
-    override fun handleConnection(conn: Connection) {
-        conn.muxerSession().createStream(listOf(this))
-    }
+    override fun handleConnection(conn: Connection) = outboundStreams.handleConnection(conn)
 
     override fun initChannel(ch: P2PChannel, selectedProtocol: String): CompletableFuture<out Unit> {
         logger.trace("Gossip initChannel - selected protocol: {}", selectedProtocol)
Index: libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/SingleOutboundStreamPerPeer.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/SingleOutboundStreamPerPeer.kt b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/SingleOutboundStreamPerPeer.kt
new file mode 100644
--- /dev/null	(date 1789054697080)
+++ b/libp2p/src/main/kotlin/io/libp2p/pubsub/gossip/SingleOutboundStreamPerPeer.kt	(date 1789054697080)
@@ -0,0 +1,83 @@
+package io.libp2p.pubsub.gossip
+
+import io.libp2p.core.Connection
+import io.libp2p.core.ConnectionHandler
+import io.libp2p.core.PeerId
+import io.libp2p.core.Stream
+import org.slf4j.LoggerFactory
+import java.util.concurrent.CompletableFuture
+
+/**
+ * Maintains a single outbound [Stream] per remote peer id, whatever number of [Connection]s
+ * there is to that peer.
+ *
+ * libp2p may hold more than one connection to the same peer - most commonly when both sides dial
+ * each other at the same time - while gossipsub, like go-libp2p-pubsub, expects a single outbound
+ * stream per peer id. Opening a stream on every connection makes both peers see same-direction
+ * duplicates and drop them, which leaves the peering permanently unable to write.
+ *
+ * [handleConnection] opens a stream when this is the first connection to the peer. Every stream
+ * opened is then watched and reopened over any connection to the peer which is still alive, be it
+ * closed together with the connection carrying it or reset by the remote peer.
+ *
+ * The [streamOpener] future is expected to complete only when the stream is ready to be used, so
+ * that a peer not supporting the protocol doesn't make us reopen the stream over and over.
+ *
+ * Thread safe.
+ */
+class SingleOutboundStreamPerPeer(
+    private val streamOpener: (Connection) -> CompletableFuture<Stream>
+) : ConnectionHandler {
+
+    private val connectionsByPeer = mutableMapOf<PeerId, MutableList<Connection>>()
+
+    override fun handleConnection(conn: Connection) {
+        val peerId = conn.remotePeerId
+        val isFirstConnection = synchronized(connectionsByPeer) {
+            val connections = connectionsByPeer.getOrPut(peerId) { mutableListOf() }
+            connections += conn
+            connections.size == 1
+        }
+        conn.closeFuture().thenRun { forgetConnection(peerId, conn) }
+        if (isFirstConnection) {
+            openStream(conn)
+        }
+    }
+
+    private fun forgetConnection(peerId: PeerId, conn: Connection) {
+        synchronized(connectionsByPeer) {
+            val connections = connectionsByPeer[peerId] ?: return
+            connections -= conn
+            if (connections.isEmpty()) {
+                connectionsByPeer -= peerId
+            }
+        }
+    }
+
+    private fun reopenStream(peerId: PeerId) {
+        val conn = aliveConnection(peerId) ?: return
+        logger.debug("Reopening the outbound stream to {}", peerId)
+        openStream(conn)
+    }
+
+    private fun aliveConnection(peerId: PeerId): Connection? =
+        synchronized(connectionsByPeer) {
+            connectionsByPeer[peerId]?.firstOrNull { !it.closeFuture().isDone }
+        }
+
+    private fun openStream(conn: Connection) {
+        streamOpener(conn).whenComplete { stream, err ->
+            if (err != null) {
+                logger.debug("Couldn't open an outbound stream to {}", conn.remotePeerId, err)
+            } else {
+                stream.closeFuture().thenRun { reopenStream(conn.remotePeerId) }
+            }
+        }
+    }
+
+    private val Connection.remotePeerId: PeerId get() = secureSession().remoteId
+
+    companion object {
+        private val logger = LoggerFactory.getLogger(SingleOutboundStreamPerPeer::class.java)
+    }
+}
```

Nashatyrev (commented):
LGTM

PR #531, opened 2026-09-16
feat(gossip): bound inbound RPC protobuf field count

lucassaldanha:
## Inbound: bound the field count

`maxControlMessageSize` bounds allocation only indirectly: the cheapest object protobuf-java materialises, an empty repeated sub-message or a retained unknown field, costs two wire bytes, so a 256 KiB budget still admits ~131k allocations from one frame. `maxTotalFields` caps that directly, defaulting to 65536 to keep at least four wire bytes per field.

`RpcMessageCountValidator` walks the frame before `ProtobufDecoder` materialises it and charges one field per occurrence at every nesting level. It descends into every field protobuf-java parses recursively:

- subscriptions, publish, control, and `partial`, which `skipField` previously swallowed whole;
- `testExtension`, a known but empty message field whose entire interior is retained as unknown fields;
- unknown groups, which protobuf-java expands field by field into a nested `UnknownFieldSet`. `END_GROUP` delimiters are not charged, since they are not materialised as fields.

Unknown length-delimited fields need no such treatment: protobuf stores the body as one opaque `ByteString` and never parses its interior.

Exceeding the limit drops the frame without scoring the peer. No other implementation bounds field count, so a peer configured more loosely than us is conformant rather than hostile. `GossipParams` validates `maxTotalFields > 0` (or null to disable).

## Outbound: keep symmetry

The batcher must never emit an RPC that a peer running this same code would reject pre-decode. A publish's `data` payload is exempt from the control-byte budget but each `data` field still costs one field inbound, so a burst of small-envelope publishes could stay within the byte budgets while overflowing the field cap.

`DefaultGossipRpcPartsQueue.takeBatch` now also splits on a `maxTotalFields` budget, using a per-part field estimate (`estimatedMaxFieldCount`) counted the same way the inbound walker counts, including retained unknown fields and groups. `addPart` rejects a single part that alone exceeds the field budget, mirroring the existing `maxGossipMessageSize` guard, so the emit-a-lone-oversized-part rule cannot break symmetry.

A symmetry test drains large batches of minimal publishes - plain, carrying unknown fields, and carrying empty/nested unknown groups - and asserts every produced RPC passes the inbound validator under default limits.

## CI

`android-actions/setup-android` was failing at the Setup Android SDK step because its default package set requests the deprecated `tools` package, which Google removed. Bumped to `v3.2.2` and pinned `packages: platform-tools`.


.github/workflows/build.yml
```diff
@@ -37,9 +37,12 @@ jobs:
         uses: gradle/gradle-build-action@v2
 
       - name: Setup Android SDK
-        uses: android-actions/setup-android@v3
+        uses: android-actions/setup-android@v3.2.2
         with:
           cmdline-tools-version: 8512546
+          # Default packages include the deprecated 'tools' package, which Google removed and
+          # sdkmanager can no longer install; keep only platform-tools.
+          packages: platform-tools
```
lucassaldanha, line 45: Had to update this to get CI running, nothing directly related to this PR

libp2p/src/test/kotlin/io/libp2p/pubsub/gossip/RpcPartsQueueSymmetryTest.kt
```diff
@@ -0,0 +1,177 @@
...
+import io.libp2p.pubsub.RpcMessageCountValidator
+import io.libp2p.pubsub.TooLargeMessageException
+import io.netty.buffer.Unpooled
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+import pubsub.pb.Rpc
+
+/**
+ * Symmetry guard: the outbound batcher must never emit an RPC that this same library's inbound
```
lucassaldanha, line 15: This was Claude's idea to ensure we never publish a message that we would reject. So we did some changes on the busines logic of the publishing as well.

tbenr: if we want to do so we need to update the KDoc's GossipParams for `maxTotalFields`, which mentions only inbound.

tbenr:
so by default we are moving from 131k to 65k. maybe the 65k limit in the lib is reasonable but from teku we may think to lower it down.

`
Ethereum message IDs are 20 bytes, so an IHAVE/IWANT ID field costs at least ~22 wire bytes. Topics are larger still. Therefore, with a 16,384 limit, the 256 KiB control-byte budget normally fires first; the field limit mainly catches empty-envelope attacks
`

tbenr:
I run this check with gpt

RPC shape | Fields/item | Charged bytes/item | Fits in 256 KiB | Fields at byte limit | Normal cap
-- | -- | -- | -- | -- | --
IWANT / IDONTWANT ID | 1 | 22 | 11,915 | ~11,917 | 5,000
IHAVE ID, shared topic | ~1 | ~22 | ~11,913 | ~11,916 | 5,000
GRAFT with topic | 2 | 42 | 6,241 | 12,483 | Number of topics
PRUNE with topic/backoff | 3 | 44 | 5,957 | 17,872 | Number of topics
PRUNE PeerInfo, ID only | 2 | ~42 | ~6,240 | ~12,484 | 16
Standard subscription | 3 | 44 | 5,957 | 17,871 | 500
Partial-enabled subscription | 5 | 48 | 5,461 | 27,305 | 500
Partial control-extension flag | 3 total | 6 total | One logical value | 3 | One
RPC.partial | 3–5 total | Metadata-dependent | One logical value | 3–5 | One
Smallest Ethereum publish | 3 | ~47 charged | ~5,577 | ~16,731 | Traffic-dependent

lucassaldanha:
> so by default we are moving from 131k to 65k. maybe the 65k limit in the lib is reasonable but from teku we may think to lower it down.

Opened https://github.com/Consensys-Incorporated/teku/issues/11341 in Teku so we config the limit when we cut a new release of jvm-libp2p.

tbenr:
Did I went too far with fable?

[pr-531-field-count-cost.md](https://github.com/user-attachments/files/32607336/pr-531-field-count-cost.md)
[AllocBench.java](https://github.com/user-attachments/files/32607339/AllocBench.java)
[Bench.java](https://github.com/user-attachments/files/32607340/Bench.java)
[PublishBench.java](https://github.com/user-attachments/files/32607341/PublishBench.java)

